PR #1, opened 2016-12-05
Feature/response writer

daiLlew:
Added common functionality for writing JSON to http response.

### DIFF
--- a/handlers/response/writer.go
+++ b/handlers/response/writer.go
@@ -1,0 +1,36 @@
+package response
+
+import (
+	"net/http"
+	"encoding/json"
+)
+
+const contentTypeHeader = "Content-Type"
+const contentTypeJSON = "application/json"
+
+// JSONEncoder interface defining a JSON encoder.
+type JSONEncoder interface {
+	writeResponseJSON(w http.ResponseWriter, value interface{}, status int) error
+}
+
+type onsJSONEncoder struct{}
+
+var jsonResponseEncoder JSONEncoder = &onsJSONEncoder{}
+
+// WriteJSON set the content type header to JSON, writes the response object as json and sets the http status code.
+func WriteJSON(w http.ResponseWriter, value interface{}, status int) error {
+	return jsonResponseEncoder.writeResponseJSON(w, value, status);
+}
+
+func (j *onsJSONEncoder) writeResponseJSON(w http.ResponseWriter, value interface{}, status int) error {
+	w.Header().Set(contentTypeHeader, contentTypeJSON)
+
+	b, err := json.Marshal(value)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return err
+	}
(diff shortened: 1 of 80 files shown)


handlers/response/writer.go
```diff
@@ -0,0 +1,36 @@
+package response
+
+import (
+	"net/http"
+	"encoding/json"
```
CarlHembrough: Run go fmt

handlers/response/writer.go
```diff
@@ -0,0 +1,36 @@
+package response
```
CarlHembrough: Consider if this package should be under the 'handlers' directory. Not sure if there are any conventions here, but currently there are only actual handlers in there. Should it only be handlers in that directory or anything handler related?

handlers/response/writer_test.go
```diff
@@ -0,0 +1,75 @@
...
+	Name  string `json:"name"`
+	Child child `json:"child"`
+}
+
+type mockOnsJsonEncoder struct {
+	invocationArgs  []interface{}
+	mockedBehaviour func(w http.ResponseWriter, value interface{}, status int) error
+}
+
+func mockEncoderInit() *mockOnsJsonEncoder {
```
CarlHembrough: go lint is saying that 'mockOnsJsonEncoder' should be mockOnsJSONEncoder. Should we even upper case 'Ons'?

handlers/response/writer_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (mock *mockOnsJsonEncoder) writeResponseJSON(w http.ResponseWriter, value interface{}, status int) error {
+	mock.invocationArgs = append(mock.invocationArgs, value)
+	return mock.mockedBehaviour(w, value, status)
+}
+
+func TestWriteJSON(t *testing.T) {
+
+	Convey("Should Write value to response as JSON, set the expected http response status code & set the content " +
+		"type header with the expected value", t, func() {
```
CarlHembrough: Consider splitting up the long single statment into multiple ''Convey" calls to improve output format.

handlers/response/writer_test.go
```diff
@@ -0,0 +1,75 @@
+package response
```
CarlHembrough: Consider putting tests in package 'response_test' to make the tests external to the package, while still in the same directory.

handlers/response/writer.go
daiLlew: I don't know. My interpretation of the go packages is the name describes what it is. This isn't a handler its a response writer... I'll ask Ian

handlers/response/writer_test.go
daiLlew: Is this the go way?

ian-kent: no reason not to do both - we can have multiple test files, one which tests the package internally and another which tests the external interface, if that adds value

PR #10, opened 2017-07-07
Add common AWS S3 reader that handles fully qualified URL's 

CarlHembrough:
# What

Add common AWS S3 reader that handles fully qualified URL's that include the bucket name.

# How to review

- check the implementation is appropriate
- suggest ways of adding test coverage to s3


s3/s3.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// Use the S3 URL implementation as the S3 drivers don't seem to handle fully qualified URLs that include the
+	// bucket name.
+	url, err := NewURL(rawURL)
+	if err != nil {
+		return nil, err
+	}
+
+	bucket := s3.s3.Bucket(url.BucketName())
+	reader, error := bucket.GetReader(url.Path())
```
eldeal: err not error?

s3/s3.go
```diff
@@ -0,0 +1,48 @@
...
+	s3 := s3.New(auth, aws.Regions[awsRegion])
+
+	return &S3{
+		s3: s3,
+	}, nil
+}
+
+// S3 provides AWS S3 functions that support fully qualified URL's.
+type S3 struct {
+	s3 *s3.S3
```
eldeal: Could this be embedded as with the URL struct?

s3/url_test.go
```diff
@@ -0,0 +1,38 @@
+package s3_test
```
eldeal: Torn on this one, I've not generally seen the package redefined within itself. And if this stayed as s3 you wouldn't need to import the s3 package. But I suppose that means the TestSpec method is exported as part of the s3 package.

This currently breaks the convention set by other tests in this repo, but may be worth a discussion around rather than assuming they're correct by default.

s3/s3.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"io"
+	"github.com/goamz/goamz/aws"
+	"github.com/goamz/goamz/s3"
+	"time"
+)
+
+// New returns a new AWS specific file.Provider instance configured for the given region.
+func New(awsRegion string) (*S3, error) {
```
CarlHembrough: Done.

s3/s3.go
```diff
@@ -0,0 +1,48 @@
...
+
+	s3 := s3.New(auth, aws.Regions[awsRegion])
+
+	return &S3{
+		s3: s3,
+	}, nil
+}
+
+// S3 provides AWS S3 functions that support fully qualified URL's.
+type S3 struct {
```
CarlHembrough: I agree - done!

s3/s3.go
CarlHembrough: Yes 👍

s3/s3.go
```diff
@@ -0,0 +1,48 @@
...
+	s3 *s3.S3
+}
+
+// Get a io.ReadCloser instance for the given fully qualified S3 URL.
```
CarlHembrough: Done.

s3/s3.go
CarlHembrough: good spot - changed.

s3/url_test.go
CarlHembrough: This is a patterned picked up here: https://medium.com/@matryer/5-simple-tips-and-tricks-for-writing-unit-tests-in-golang-619653f90742

It suggests putting tests in a separate package to hide the internals, and test via the public interface of the package. Happy to discuss and change if we decide on another convention.

eldeal: That makes a lot of sense 👍

PR #11, opened 2017-07-10
Add zebedee client

mattrout92:
### What

-Add a zebedee client to go-ns
-Unit tests for zebedee client
-Add zebedee page models

### How to review

-Verify the page models are correct
-Create a basic service which uses the client to test zebedee responses

### Who can review

Anyone


zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+
+	fs := struct {
+		Size int `json:"fileSize"`
+	}{}
+	if err = json.Unmarshal(b, &fs); err != nil {
+		log.Error(err, nil)
+		return ""
+	}
+
+	return datasize.ByteSize(fs.Size).HumanReadable()
```
ian-kent: feels like the wrong place to be formatting the result of a filesize - I'd expect the zebedee client to return an int64 or something containing the number of bytes, it should only be converted to a human readable string in the view

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+// https://github.com/ONSdigital/zebedee
+//
+// Get will return a response from the server as a byte array, usually in a json
+// format.
+//
+// GetLanding will return a static dataset landing page struct populated with the
+// data from a given path in zebedee.
+type Client interface {
+	Get(path string) ([]byte, error)
+	GetLanding(path string) (StaticDatasetLandingPage, error)
```
ian-kent: could be more explicit name, e.g. `GetDatasetLandingPage`?

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+
+// Client is an interface for communication to a zebedee server:
+// https://github.com/ONSdigital/zebedee
+//
+// Get will return a response from the server as a byte array, usually in a json
+// format.
+//
+// GetLanding will return a static dataset landing page struct populated with the
+// data from a given path in zebedee.
+type Client interface {
```
ian-kent: also based on `interfaces in, structs out` do we even need to define the Client interface here?

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+type StaticDatasetLandingPage datasetLandingPageStatic.Page
+
+// ErrInvalidZebedeeResponse is returned when zebedee does not respond
+// with a valid status
+type ErrInvalidZebedeeResponse struct {
+	err  error
+	data zebedeeRequestErrorData
+}
+
+type zebedeeRequestErrorData map[string]interface{}
```
ian-kent: could this be a struct rather than a map? and it could be embedded directly into `ErrInvalidZebedeeResponse`

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+
+func (e ErrInvalidZebedeeResponse) Error() string {
+	return e.err.Error()
```
ian-kent: i'd consider stringifying the `data` field too so it shows up in logs, possibly as JSON, or possibly with `%+v` in sprintf

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+		zebedeeURL: url,
+		client:     &http.Client{Timeout: 5 * time.Second},
+	}
+}
+
+func (c ZebedeeClient) Get(path string) ([]byte, error) {
+	return c.get(path)
+}
+
+func (c ZebedeeClient) GetLanding(path string) (StaticDatasetLandingPage, error) {
```
ian-kent: same as on interface, name could be `GetDatasetLandingPage`

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+			}(i, e, element)
+		}
+	}
+	wg.Wait()
+
+	return c.populateStaticLandingPageModel(*dlp)
+}
+
+func (c ZebedeeClient) SetAccessToken(token string) {
+	log.Debug("adding access token to client", log.Data{"token": token})
```
ian-kent: we shouldn't be logging out the access token, it's equivalent to a password (actually probably more secure, since with the access token you don't need to know which email address its for)

could consider logging out the last few chars or something for context if its worthwhile?

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, ErrInvalidZebedeeResponse{errors.New("unexpected response from zebedee"), zebedeeRequestErrorData{
```
ian-kent: should drain the response body here, something like `io.Copy(ioutil.Discard, resp.Body)`

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+	sdlp.URI = dlp.URI
+	sdlp.Metadata.Title = dlp.Description.Title
+	sdlp.Metadata.Description = dlp.Description.Summary
+	sdlp.DatasetLandingPage.Related.Datasets = dlp.RelatedDatasets
+	sdlp.DatasetLandingPage.Related.Publications = dlp.RelatedDocuments
+	sdlp.DatasetLandingPage.Related.Methodology = append(dlp.RelatedMethodology, dlp.RelatedMethodologyArticle...)
+	sdlp.DatasetLandingPage.Related.Links = dlp.RelatedLinks
+	sdlp.DatasetLandingPage.IsNationalStatistic = dlp.Description.NationalStatistic
+	sdlp.DatasetLandingPage.IsTimeseries = dlp.Timeseries
+	sdlp.ContactDetails.Email = dlp.Description.Contact.Email
```
ian-kent: might be able to clean this up by wrapping some structs - if the source and dest struct have the same field names, Go 1.8+ can map it automatically, e.g. `sdlp.ContactDetails = Contact{dlp.Description.Contact}`

ian-kent: (or something like that - never used it, just seen it in the release notes)

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+	sdlp.ContactDetails.Telephone = dlp.Description.Contact.Telephone
+	sdlp.ContactDetails.Name = dlp.Description.Contact.Name
+	sdlp.DatasetLandingPage.ReleaseDate = dlp.Description.ReleaseDate
+	sdlp.DatasetLandingPage.NextRelease = dlp.Description.NextRelease
+	sdlp.DatasetLandingPage.Notes = dlp.Section.Markdown
+	sdlp.FilterID = dlp.FilterID
+	bc, err := c.getBreadcrumb(dlp.URI)
+	if err != nil {
+		return StaticDatasetLandingPage{}, err
```
ian-kent: could consider just returning `sdlp` rather than throwing away existing data, since we've done the work to get it? and mention in the godoc that if an error is returned you may only get partial data?

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	sdlp.Page.Breadcrumb = bc
+	sdlp.DatasetLandingPage.ParentPath = sdlp.Breadcrumb[len(sdlp.Breadcrumb)-1].Title
```
ian-kent: can `len(sdlp.Breadcrumb)` ever return 0? if so, should wrap this in an `if`

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+		sdlp.DatasetLandingPage.Datasets = append(sdlp.DatasetLandingPage.Datasets, dataset)
+	}
+
+	for _, value := range dlp.Alerts {
+		switch value.Type {
+		default:
+			log.Debug("Unrecognised alert type", log.Data{"alert": value})
+			fallthrough
+		case "alert":
+			sdlp.DatasetLandingPage.Notices = append(sdlp.DatasetLandingPage.Notices, datasetLandingPageStatic.Message{
```
ian-kent: same here (and a few lines below), might be able to avoid manually mapping the fields, if Date/Markdown are of the same types and there aren't additional fields which could be accidentally copied

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+	b, err := c.get("/data?uri=" + uri)
+	if err != nil {
+		log.Error(err, nil)
+		return datasetLandingPageStatic.Dataset{URI: uri}
+	}
+
+	var d models.Dataset
+	if err = json.Unmarshal(b, &d); err != nil {
+		log.Error(err, nil)
+		return datasetLandingPageStatic.Dataset{
```
ian-kent: inconsistent formatting with line 224 (don't mind which, just consistency)

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	pt := struct {
+		Title   string `json:"title"`
+		Edition string `json:"edition"`
+	}{}
+
+	if err = json.Unmarshal(b, &pt); err != nil {
+		log.Error(err, nil)
+		return ""
```
ian-kent: a json unmarshal error doesn't necessarily mean we haven't got the title/edition, so possibly worth just falling through to the block below (log the error, then see what happens)

zebedee/client.go
```diff
@@ -0,0 +1,302 @@
...
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/ONSdigital/dp-frontend-models/model"
```
ian-kent: this is a slightly misplaced comment, took a while to notice...

we should probably split the "zebedee client" and "preparing data for the frontend renderer" into two separate packages

there are probably a lot of use cases where we want to get data from zebedee (e.g. the router), but don't want it automatically mapped to the frontend models since it'll never get to the renderer

splitting into two packages (client, and mapping to renderer models) means callers only need to import dependencies which are absolutely necessary

zebedee/client/client.go
```diff
@@ -0,0 +1,201 @@
...
+	)
+}
+
+var _ error = ErrInvalidZebedeeResponse{}
+
+// NewZebedeeClient creates a new Zebedee Client
+func NewZebedeeClient(url string) ZebedeeClient {
+	return ZebedeeClient{
+		zebedeeURL: url,
+		client:     &http.Client{Timeout: 5 * time.Second},
```
ian-kent: might want to make this configurable? some zebedee responses can be _very_ slow

zebedee/client/client.go
```diff
@@ -0,0 +1,201 @@
...
+		req.Header.Set("X-Florence-Token", c.accessToken)
+	}
+
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
```
ian-kent: might be better to check for a range, e.g. `201` or `202` would be valid responses

zebedee/client/client.go
```diff
@@ -0,0 +1,201 @@
...
+		io.Copy(ioutil.Discard, resp.Body)
+		return nil, ErrInvalidZebedeeResponse{resp.StatusCode, http.StatusOK, req.URL.Path}
+	}
+
+	return ioutil.ReadAll(resp.Body)
+}
+
+// GetBreadcrumb returns a Breadcrumb
+func (c ZebedeeClient) GetBreadcrumb(uri string) ([]data.Breadcrumb, error) {
+	b, err := c.get("/parents?uri=" + uri)
```
ian-kent: don't think this syntax works? iirc we have to use `http.NewRequest` and set `req.URL.RawQuery`?

ian-kent: (same for the others in this package)

mattrout92: Not sure why this syntax wouldn't work? Certainly works when I integrate it with zebedee?

PR #12, opened 2017-07-14
Add a form validation library

mattrout92:
### What

- An extendable server side validator for validating HTML form fields

### How to review

- Read the readme to get an understanding of how to use the library
- Create your own form, and validate it using the existing rules, ensuring you handle errors appropriately

### Who can review

- Anyone

### DIFF
--- a/validator/README.md
+++ b/validator/README.md
@@ -74,13 +74,13 @@ function and handle any errors as follows:
 ```go
   var frm Form
   if err := fv.Validate(req, &frm); if err != nil {
-    if err != validator.ErrFormValidationFailed {
+    if _, ok := err.(ErrFormValidationFailed); !ok {
       // handle this error as you would with any normal error
     } else {
       // we now know we have field validation errors so we can work out why a
       // particular field failed validation
 
-      fieldErrs := fv.GetFieldErrors() // returns a map of []errors
+      fieldErrs := err.(ErrFormValidationFailed).GetFieldErrors() // returns a map of []errors
       if len(fieldErrs["search"]) > 0 {
         // we now know something with search failed validation so we can handle
         // the error appropriately, perhaps by displaying an warning message to
--- a/validator/README.md
+++ b/validator/README.md
@@ -42,13 +42,19 @@ variable within in rule is an optional field which can be passed into a validati
 function such as the one above for the min_length rule. To define your own custom
 rules please see the section further down.
 
-To use the validator you should create a new validator, passing in a path to your
-previously defined JSON file.
+To use the validator you should create a new validator, passing in an `io.Reader`
+with the contents of your JSON rules. For example:
 
 ```go
 import "github.com/ONSdigital/go-ns/validator"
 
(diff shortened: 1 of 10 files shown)


validator/README.md
```diff
@@ -0,0 +1,152 @@
...
+to define validation logic of form fields in a JSON file to allow potential
+validation sharing with the client side.
+
+#### How to use the validator
+
+Imagine you had an HTML form as described below:
+
+```html
+<form action="/search" method="post">
+  <input type="text" id="search" value="search">
```
ian-kent: html form would use `name` not `id` for submission

validator/README.md
```diff
@@ -0,0 +1,152 @@
...
+      },
+      {
+        "name": "email"
+      }
+    ]
+  }
+]
+```
+
+The id field would match the input id from the HTML, and we then define a list of
```
ian-kent: same here, it's the input `name` which is used not the id

validator/rules.go
```diff
@@ -0,0 +1,99 @@
...
+
+func email(vars ...interface{}) error {
+	var e string
+	var ok bool
+
+	if e, ok = vars[0].(string); !ok {
+		return errors.New("first parameter to email must be a string")
+	}
+
+	if ok, err := regexp.MatchString(".+@.+\\..+", e); !ok || err != nil {
```
ian-kent: could use a more specific regex? http://emailregex.com/

current regex would allow invalid email addresses e.g. `user@user@user@user.com`

validator/validate.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/gorilla/schema"
+)
+
+// ErrFormValidationFailed if form validation has failed, if this is ever returned,
+// corresponding call to GetFieldErrors should be made to check for individual field errors
+var ErrFormValidationFailed = errors.New("form validation failed, check field errors")
+
+// FormValidator represents a form validator with appropriate validator method
+type FormValidator struct {
+	rulesJSON []byte
```
ian-kent: we should consider how we could make this goroutine-safe? i.e. so `FormValidator` can be created once with the rules JSON and reused across multiple forms/requests?

currently it'd require parsing the rules JSON on every request

validator/validate.go
```diff
@@ -0,0 +1,103 @@
...
+
+// ErrFormValidationFailed if form validation has failed, if this is ever returned,
+// corresponding call to GetFieldErrors should be made to check for individual field errors
+var ErrFormValidationFailed = errors.New("form validation failed, check field errors")
+
+// FormValidator represents a form validator which can be used to validate an
+// HTML form
+type FormValidator struct {
+	flds      []field
+	fieldErrs map[string][]error
```
ian-kent: this bit needs to change to be goroutine safe

i.e., when validation happens, a new instance of this needs to be created and that gets returned by the validate call, otherwise atm multiple goroutines using the same instance will be overwriting/modifying each others field errors

PR #13, opened 2017-07-18
Feature/kafka add

gedge:


kafka/consumer-group.go
```diff
@@ -0,0 +1,81 @@
...
+			case err := <-cg.Consumer.Errors():
+				log.Error(err, nil)
+				cg.Errors <- err
+			default:
+				select {
+				case msg := <-cg.Consumer.Messages():
+					cg.Incoming <- Message{msg, cg.Consumer}
+				case n, more := <-cg.Consumer.Notifications():
+					if more {
+						log.Trace("Rebalancing group", log.Data{"topic": topic, "group": group, "partitions": n.Current[topic]})
```
nshumoogum: Generally Im not sure we want to be logging out messages in our library as it forces the public to use `go-ns/log` which might not be what they desire.

For this case if we are not sending this information back up stream (via a channel) do we really want to be logging out the information? Another way of looking at it is a user of this library cannot declare whether they care about the consumer group rebalancing, so why force this log line on them??

kafka/consumer.go
```diff
@@ -0,0 +1,51 @@
...
+
+	go func() {
+		defer consumer.Close()
+		log.Info("Started kafka consumer", log.Data{"topic": topic})
+		for {
+			select {
+			case err := <-consumer.Errors():
+				log.Error(err, log.Data{"topic": topic})
+				return
+			default:
```
nshumoogum: Is it necessary to set a default case here which encapsulates another switch statement?

nshumoogum:
As stated in one of my comments, generally do we want to be logging out messages in our library as it forces the public to use `go-ns/log` which might not be what they desire?

kafka/consumer-group.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/bsm/sarama-cluster"
+)
+
+var tick = time.Millisecond * 4000
+
+type ConsumerGroup struct {
+	Consumer *cluster.Consumer
+	Incoming chan Message
+	Closer   chan bool
+	Errors   chan error
```
CarlHembrough: In the observation extractor I ended up adding methods to return 'Incoming' and 'Closer' and making the struct fields private. This enabled me to create an interface like this 

```
// MessageConsumer provides a generic interface for consuming []byte messages
type MessageConsumer interface {
	Incoming() chan kafka.Message
	Closer() chan bool
}
```
I could then easily test around Kafka using a mock consumer: https://github.com/ONSdigital/dp-observation-extractor/blob/feature/initial-implementation/event/eventtest/message_consumer.go

Do you think its worth having the struct members private with methods to access them? or is there another way of mocking them that you can show me? (this also applied to the producer)

kafka/consumer-group.go
```diff
@@ -0,0 +1,81 @@
...
+	consumer *cluster.Consumer
+}
+
+func (M Message) GetData() []byte {
+	return M.message.Value
+}
+
+func (M Message) Commit() {
+	M.consumer.MarkOffset(M.message, "metadata")
+	//M.consumer.CommitOffsets()
```
gedge: 👍

kafka/consumer-group.go
gedge: I agree that a library should _generally_ not force logging decisions on you.

The kafka package was created to capture a common usage pattern across the pipeline's (and now also perceived `go-ns` users') microservices, and hence replicated that user need.

Given that it was meant to be a thin abstraction layer, and that you've pointed out (verbally) that it doesn't really add much value in general, this is perhaps one place where value is added (albeit using a forced/controversial design decision!) by taking care of the monitoring and logging of re-partitioning.

I guess I'm saying that if you want to avoid that "added" value, then don't use this library. But you're already using `go-ns` so use our logging, too.

kafka/consumer.go
gedge: As I recall, we had a problem when soak-testing a flat `select` statement.  The busy `consumer.Messages()` always took precedence over the errors channel, so this multi-layer version forced the priority of error handling over message consumption.

kafka/consumer-group.go
gedge: Perhaps a compromise would be to use an interface (`Logger` would have been nice) and pass in the logger object, but that would involve a lot more reworking of the `log` lib.

kafka/consumer-group.go
gedge: how does this look: https://github.com/ONSdigital/go-ns/pull/14

kafka/consumer.go
nshumoogum: Okay fair enough

nshumoogum:
👍

PR #20, opened 2017-08-29
Move out clients

mattrout92:
### What

Moved out clients to go-ns as they are not specific to the filter-dataset-controller

### How to review

Start up usual frontend services, including hierarchy-api, filter-api, code-list-api
Verify that the filter user journey has not broken by going to a dataset landing page and navigating the journey

### Who can review

Anyone


clients/dataset/dataset.go
```diff
@@ -0,0 +1,78 @@
...
+
+// ErrInvalidDatasetAPIResponse is returned when the dataset api does not respond
+// with a valid status
+type ErrInvalidDatasetAPIResponse struct {
+	expectedCode int
+	actualCode   int
+	uri          string
+}
+
+func (e ErrInvalidDatasetAPIResponse) Error() string {
```
eldeal: Comment :)

clients/dataset/dataset.go
```diff
@@ -0,0 +1,78 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return "dataset-api", &ErrInvalidDatasetAPIResponse{http.StatusOK, resp.StatusCode, "/healthcheck"}
+	}
+
+	return "", nil
+}
+
+// GetDataset ...
+func (c *Client) GetDataset(id, edition, version string) (m Model, err error) {
```
eldeal: Probably shouldn't call it GetDataset when it's actually fetching the Version - could cause confusion when the aim is to actually retrieve dataset metadata

clients/hierarchy/data.go
```diff
@@ -0,0 +1,26 @@
...
+	Children []Child `json:"children"`
+	Parent   Parent  `json:"parent"`
+}
+
+// Child represents a child item in the hierarchy model
+type Child struct {
+	ID               string `json:"id"`
+	Label            string `json:"label"`
+	URL              string `json:"url"`
+	NumberofChildren int    `json:"number_of_children"`
```
eldeal: I think there should be some omitempty's in these structs - number_of_children for one, possibly Parent and Children on the Model struct?

clients/renderer/renderer.go
```diff
@@ -0,0 +1,74 @@
+package renderer
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+// ErrInvalidRendererResponse is returned when the renderer service does not respons
```
eldeal: Typo in comment - respond not respons

zebedee/client/client.go
```diff
@@ -39,25 +39,25 @@ var _ error = ErrInvalidZebedeeResponse{}
...
 
 // Get returns a response for the requested uri in zebedee
-func (c ZebedeeClient) Get(path string) ([]byte, error) {
+func (c *ZebedeeClient) Get(path string) ([]byte, error) {
 	return c.get(path)
 }
 
 // Healthcheck calls the healthcheck endpoint on the api and alerts the caller of any errors
-func (c *ZebedeeClient) Healthcheck() (string, error) {
+func (c ZebedeeClient) Healthcheck() (string, error) {
```
eldeal: Why not a pointer here but in all other healthchecks and other methods on this client?

clients/filter/data.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// DimensionOption represents a dimension option from the filter api
+type DimensionOption struct {
+	DimensionOptionsURL string `json:"dimension_option_url"`
+	Option              string `json:"option"`
+}
+
+// Model represents a model returned from the filter api
+type Model struct {
```
eldeal: Possibly need some omitempty's as the filter can be posted in bits and pieces - downloads might not be passed in initially, events might only get created one it's submitted and the system takes over

clients/filter/data.go
```diff
@@ -0,0 +1,52 @@
...
+	Downloads       map[string]Download `json:"downloads"`
+	Events          map[string][]Event  `json:"events"`
+}
+
+// ModelDimension represents a dimension to be filtered upon
+type ModelDimension struct {
+	Name      string    `json:"name"`
+	Values    []string  `json:"values"`
+	IDs       []string  `json:"ids"`
+	Hierarchy Hierarchy `json:"hierarchy"`
```
eldeal: How does this deal with codelists when the dimension is not a hierarchy?

mattrout92:
Removed the hierarchy struct on the filter dimensions client, as it is no longer used anywhere in the code base.

eldeal:
This branch contains the rhttp package, which has been reviewed separately in another PR. As long as the correct version gets to master I'm happy with the other changes in this PR

PR #21, opened 2017-08-29
Feature/configurable signal handling

CarlHembrough:
The go-ns http server is handling OS signals, and shutting down the server when a signal is received. This is useful for HTTP only services, but other services require more control over what happens when a signal is received. The default behaviour also prevents a context from being passed to the shutdown function on the server. To cater for this I have allowed the default handling of OS signals to be disabled. Still providing it as a default without changing the interface should not break backwards compatibility.

 - Allow OS signal handling to be disabled, allowing the shutdown to be handled externally.
 - Provide a default context with a timeout when using the default OS signal handling.
 - Allow a default shutdown time to be provided which is used when creating contexts when one is not provided.
 - Create a context using the default shutdown time if one is not passed to shutdown method.
 - Add missing vendor dependencies.


server/server.go
```diff
@@ -69,36 +74,74 @@ func (s *Server) prep() {
...
+
+	return s.Server.ListenAndServe()
+}
+
+func (s *Server) listenAndServeHandleOSSignals() error {
 	stop := make(chan os.Signal, 1)
 	signal.Notify(stop, os.Interrupt, os.Kill)
 
+	s.prep()
 	if len(s.CertFile) > 0 || len(s.KeyFile) > 0 {
```
ian-kent: could just refactor this bit to move it into `listenAndServeAsync`, and get rid of `listenAndServeAsyncTLS`?

server/server.go
```diff
@@ -69,36 +74,74 @@ func (s *Server) prep() {
...
+func (s *Server) listenAndServeAsync() {
+	go func() {
+		if err := s.Server.ListenAndServe(); err != nil {
+			log.Error(err, nil)
+			os.Exit(1)
+		}
+	}()
+}
+
+func (s *Server) listenAndServeTLSAsync(certFile, keyFile string) {
```
ian-kent: `certFile` and `keyFile` parameters aren't used (using the ones on `s` is probably right)

CarlHembrough: You're right, these parameters were left over by mistake when refactoring. I have removed them.

server/server.go
CarlHembrough: Good point. I have made the change and it looks much cleaner.

PR #22, opened 2017-08-30
Change healthcheck to RWMutex and resolve concurrently

mattrout92:
- Update mutex to be RW
- Resolve external healthchecks concurrently

### DIFF
--- a/healthcheck/healthcheck.go
+++ b/healthcheck/healthcheck.go
@@ -23,30 +23,44 @@ var (
 func MonitorExternal(clients ...Client) {
 	hs := make(map[string]error)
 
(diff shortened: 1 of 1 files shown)


healthcheck/healthcheck.go
```diff
@@ -15,21 +15,33 @@ type Client interface {
...
+				mutex.Unlock()
+			}
+			wg.Done()
+		}(client)
 	}
-	mutex.Unlock()
+
+	wg.Wait()
+
+	mutex.RLock()
```
ian-kent: `RLock` and `Lock` are backwards - `R` variants are for reads

PR #23, opened 2017-09-01
Add close function to kafka consumer / producer. 

CarlHembrough:
Add close function to kafka consumer / producer. The function will block until the close operation is complete.

Previously a close would be initiated via the closer channel. The problem was that the application could exit before the graceful close had completed.


kafka/consumer-group.go
```diff
@@ -7,52 +7,45 @@ import (
...
 func (cg ConsumerGroup) Errors() chan error {
 	return cg.errors
 }
 
-func (cg ConsumerGroup) Closer() chan bool {
-	return cg.closer
+// Close safely closes the consumer and releases all resources
+func (cg *ConsumerGroup) Close() (err error) {
+
+	cg.closer <- true
```
daiLlew: Do you need to increment the wg count for each task you want to wait on?

kafka/consumer-group.go
```diff
@@ -64,32 +57,36 @@ func NewConsumerGroup(brokers []string, topic string, group string, offset int64
 		return nil, fmt.Errorf("Bad NewConsumer of %q: %s", topic, err)
 	}
 
+	var wg sync.WaitGroup
+	wg.Add(1)
+
```
daiLlew: Nevermind.

kafka/consumer-group.go
```diff
@@ -5,54 +5,46 @@ import (
...
 
 var tick = time.Millisecond * 4000
 
+// ConsumerGroup represents a Kafka consumer group instance.
 type ConsumerGroup struct {
 	consumer *cluster.Consumer
 	incoming chan Message
-	closer   chan bool
 	errors   chan error
+	closer   chan bool
```
Matt-Guest: Is there a need for the close channel if there a function that does it?

CarlHembrough: I think so - the closer channel is now used internally to signal the go routine to exit. Open to suggestions of alternatives 👍

cmd/kafka-example/main.go
```diff
@@ -56,41 +57,40 @@ func main() {
...
-				log.Info("Quitting after signal", nil)
-				producer.Closer() <- true
-				consumer.Closer() <- true
-				exitChannel <- true
-				return
-			}
+	shutdownGracefully := func() {
+
+		// give the app 3 seconds to close gracefully before killing it.
+		ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
daiLlew: Minor comment - but do we want to make the timeout configurable?

CarlHembrough: Probably not for the example, but definitely in production apps

kafka/message.go
```diff
@@ -0,0 +1,24 @@
+package kafka
+
+import (
+	"github.com/Shopify/sarama"
+	"github.com/bsm/sarama-cluster"
+)
+
+type Message interface {
```
Matt-Guest: Should there be godocs for the interface / struct and the functions?

CarlHembrough: Done

PR #24, opened 2017-09-04
Add a robust http package

mattrout92:
### What
- Added a robust http package
- Implements standard http client with additon of timeouts and
exponential backoff
- Gives the option of either using a default client (like http)
or configuring your own robust client.

###How to review
- Use the rhttp default client to send a request to a server

### Who can review
Anyone

### DIFF
--- a/rhttp/client.go
+++ b/rhttp/client.go
@@ -8,8 +8,6 @@ import (
 	"net/http"
 	"net/url"
 	"time"
-
-	"github.com/ONSdigital/go-ns/log"
 )
 
 // Client is an extension of the net/http client with ability to add
@@ -52,7 +50,7 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 		return nil, err
 	}
 
-	if resp.StatusCode >= http.StatusInternalServerError {
+	if resp.StatusCode >= http.StatusInternalServerError && c.ExponentialBackoff {
 		return c.doWithBackoff(req)
 	}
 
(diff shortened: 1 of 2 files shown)


rhttp/client.go
```diff
@@ -0,0 +1,233 @@
...
+
+	if resp.StatusCode >= http.StatusInternalServerError && c.ExponentialBackoff {
+		return c.doWithBackoff(req)
+	}
+
+	return resp, err
+}
+
+func (c *Client) doWithBackoff(req *http.Request) (resp *http.Response, err error) {
+	for i := 0; i < c.MaxRetries; i++ {
```
eldeal: if this was `i := 1; i <= c.MaxRetries` you wouldn't need to create the attempt var

rhttp/client.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	if resp.StatusCode >= http.StatusInternalServerError && c.ExponentialBackoff {
+		return c.getWithBackoff(url)
+	}
+
+	return resp, err
+}
+
+func (c *Client) getWithBackoff(url string) (resp *http.Response, err error) {
```
eldeal: your with backoff methods could probably be refactored to take the function as an argument so you dont have to repeat the for loop etc, though the differing arguments  each takes might mean you can only actually share the method for one or two uses so may not be so valuable. Just a thought :)

rhttp/client.go
```diff
@@ -0,0 +1,233 @@
...
+			}
+			continue
+		}
+
+		return
+	}
+	return
+}
+
+func getSleepTime(attempt int, retryTime time.Duration) time.Duration {
```
eldeal: A small explanatory comment for this algorithm might be helpful

PR #26, opened 2017-09-07
Remove broken timeout handler

CarlHembrough:
Remove timeout handler as suggested by @ian-kent 

- Remove broken timeout handler. 
- Remove timeout handler from default HTTP server middleware. 
- Remove unused vendor files.


server/server.go
```diff
@@ -32,13 +31,12 @@ func New(bindAddr string, router http.Handler) *Server {
...
 		"RequestID": requestID.Handler(16),
 		"Log":       log.Handler,
-		"Timeout":   timeout.Handler(10 * time.Second),
 	}
 
 	return &Server{
 		Alice:           nil,
 		Middleware:      middleware,
-		MiddlewareOrder: []string{"RequestID", "Log", "Timeout"},
+		MiddlewareOrder: []string{"RequestID", "Log"},
```
Matt-Guest: Could both "RequestID" and Log" be constants?

CarlHembrough: Done

PR #28, opened 2017-09-21
rhttp client with (cancellable) context

gedge:


cmd/rchttp-example/main.go
```diff
@@ -0,0 +1,58 @@
+package main
+
+import (
+	"context"
+	// "errors"
```
mattrout92: Can probably remove this line

PR #30, opened 2017-09-22
Create mongo package to close sessions

nshumoogum:
mongo package includes close method to close mongo sessions on service
shutdown.

### DIFF
--- a/mongo/mongo.go
+++ b/mongo/mongo.go
@@ -1,0 +1,50 @@
+package mongo
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	mgo "gopkg.in/mgo.v2"
+)
+
+// Mongo represents a simplistic MongoDB configuration.
+type Mongo struct {
+	Collection string
+	Database   string
+	URI        string
+}
+
+func (m *Mongo) Close(ctx context.Context, session *mgo.Session) error {
+	closedChannel := make(chan bool)
+	defer close(closedChannel)
+
+	timeLeft := 1000 * time.Millisecond
+	if deadline, ok := ctx.Deadline(); ok {
+		// Add some time to timeLeft so case where ctx.Done in select
+		// statement below gets called before time.After(timeLeft) gets called.
+		// This is so the context error is returned over hardcoded error.
+		timeLeft = deadline.Sub(time.Now()) + (10 * time.Millisecond)
+	}
+
+	go func() {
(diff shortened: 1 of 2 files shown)


mongo/mongo.go
```diff
@@ -0,0 +1,43 @@
...
+	"time"
+
+	mgo "gopkg.in/mgo.v2"
+)
+
+func Close(ctx context.Context, session *mgo.Session) error {
+	closedChannel := make(chan bool)
+	defer close(closedChannel)
+
+	timeLeft := 1000 * time.Millisecond
```
Matt-Guest: Is it better to have the timeout as a parameter to the function? As it allows the user of the library to control it.

mongo/mongo.go
```diff
@@ -0,0 +1,43 @@
...
+		// This is so the context error is returned over hardcoded error.
+		timeLeft = deadline.Sub(time.Now()) + (10 * time.Millisecond)
+	}
+
+	go func() {
+		// Uncomment the four line below for unit tests
+		/*time.Sleep(1100 * time.Millisecond)
+		if ctx.Value("return") == "true" || ctx.Err() != nil {
+			return
+		}*/
```
Matt-Guest: Is there any way to move this test code out?

mongo/mongo.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	mgo "gopkg.in/mgo.v2"
+)
+
+func Close(ctx context.Context, session *mgo.Session) error {
```
Matt-Guest: Missing Go docs

mongo/mongo.go
nshumoogum: Not really, if a user wants to control how the sessions close they should use context deadline. The timeLeft variable is a default value just in case a deadline is not set. By giving them control of timeLeft could lead a user to using this variable over the context deadline defeating  the purpose of context.

Instead i will set `timeLeft` variable to a constant.

mongo/mongo.go
```diff
@@ -8,11 +8,30 @@ import (
 	mgo "gopkg.in/mgo.v2"
 )
 
+// Type represents an interface to the shutdown method
+type Type interface {
```
Matt-Guest: What about `Shutdown` or `Closable`?

PR #34, opened 2017-10-09
Add logging to client packages

mattrout92:
Added clientlog package which should be used just before a client makes a request to its associated service

### DIFF
--- a/clients/filter/filter.go
+++ b/clients/filter/filter.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 
 	"github.com/ONSdigital/go-ns/clients/clientlog"
+	"github.com/ONSdigital/go-ns/log"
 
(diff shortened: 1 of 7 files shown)


clients/filter/filter.go
```diff
@@ -143,7 +156,10 @@ func (c *Client) CreateJob(datasetFilterID string) (string, error) {
 		return "", err
 	}
 
-	resp, err := c.cli.Post(c.url+"/filters", "application/json", bytes.NewBuffer(b))
+	uri := c.url + "/filters"
+	clientlog.Do(fmt.Sprintf("attemping to create filter job with instance id: %s", datasetFilterID), service, uri, "POST")
```
eldeal: if Do accepted an optional `log.Data` param, for post/put requests the relevant fields could be passed through in a structured way?

PR #35, opened 2017-10-12
tidy up all those channels by using a context instead

gedge:
this is what I was thinking


gedge:
resolved conflicts

gedge:
good to go?

PR #36, opened 2017-10-17
Feature/robust kafka 3

gedge:
more improvements, including
- using `envconfig`
- adding `NewSyncConsumer` with `CommitAndRelease`

### DIFF
--- a/cmd/kafka-example/main.go
+++ b/cmd/kafka-example/main.go
@@ -24,6 +24,8 @@ type Config struct {
 	KafkaSync     bool          `envconfig:"KAFKA_SYNC"`
 	TimeOut       time.Duration `envconfig:"GRACEFUL_SHUTDOWN_TIMEOUT"`
 	Chomp         bool          `envconfig:"CHOMP_MSG"`
+	Snooze        bool          `envconfig:"SNOOZE"`
+	OverSleep     bool          `envconfig:"OVERSLEEP"`
 }
 
 func main() {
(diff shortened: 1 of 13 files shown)


cmd/kafka-example/main.go
```diff
@@ -55,80 +67,112 @@ func main() {
...
 			case <-eventLoopContext.Done():
+				log.Trace("[KAFKA-TEST] Event loop context done", log.Data{"eventLoopContextErr": eventLoopContext.Err()})
 				return
 			case consumedMessage := <-consumer.Incoming():
-				log.Info("[KAFKA-TEST] Received message", nil)
+				consumeCount++
+				log.Info("[KAFKA-TEST] Received message", log.Data{"consumeCount": consumeCount, "consumeMax": cfg.ConsumeMax})
 				consumedData := consumedMessage.GetData()
+				sleep := 0 * time.Millisecond
+				// sleep = 500*time.Millisecond + cfg.TimeOut
```
nshumoogum: I don't know whether its worth leaving a comment to explain why you would switch out sleep of 0 seconds to the timeOut value plus half a second?

kafka/consumer-group.go
```diff
@@ -32,25 +34,38 @@ func (cg ConsumerGroup) Errors() chan error {
...
 
-	var q struct{}
-	cg.closer <- q
+	close(cg.closer)
 
 	select {
 	case <-cg.closed:
-		log.Info("Stopped listening to kafka consumer group", log.Data{"topic": cg.topic})
-		return
+		log.Info("StopListeningToConsumer saw closed kafka consumer", log.Data{"topic": cg.topic, "group": cg.group})
```
nshumoogum: The message doesn't make sense to me, Im not sure what the problem was with the previous log message?

gedge: I changed the comment again

PR #42, opened 2017-11-21
Fix/post retry payload

gedge:
any (e.g. timeout causing a) retry on a POST causes the retry to fall over (could not re-read the Body content)

### DIFF
--- a/rchttp/client.go
+++ b/rchttp/client.go
@@ -110,6 +110,17 @@ func (c *Client) Post(ctx context.Context, url string, contentType string, body 
 	return c.Do(ctx, req)
 }
 
+// Put calls Do with a PUT and the appropriate content-type and body
+func (c *Client) Put(ctx context.Context, url string, contentType string, body io.Reader) (*http.Response, error) {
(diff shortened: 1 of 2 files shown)


rchttp/client.go
```diff
@@ -43,10 +43,25 @@ var DefaultClient = &Client{
...
+	c = DefaultClient
+	c.HTTPClient.Timeout = timeout
+	return c
+}
+
 // Do calls ctxhttp.Do with the addition of exponential backoff
 func (c *Client) Do(ctx context.Context, req *http.Request) (*http.Response, error) {
 	doer := func(args ...interface{}) (*http.Response, error) {
-		return ctxhttp.Do(args[0].(context.Context), args[1].(*http.Client), args[2].(*http.Request))
+		req := args[2].(*http.Request)
```
Matt-Guest: Should the length of the array and the cast be checked?

gedge: it's not an exported function, so I imagine it wasn't necessary to check internal usage

PR #51, opened 2017-12-11
removed put version downloads & replaced with put version to match latest version of dataset api

daiLlew:
Removed PUT version downloads from dataset api client & replaced with PUT version to match latest version of dataset api spec.


clients/dataset/dataset.go
```diff
@@ -262,12 +274,12 @@ func (c *Client) GetVersion(id, edition, version string) (m Version, err error)
 	return
 }
 
-// PutVersionDownloads add the given downloads to the specified dataset version
-func (c *Client) PutVersionDownloads(datasetID string, edition string, version string, downloadList map[string]Download) error {
-	uri := fmt.Sprintf("%s/datasets/%s/editions/%s/versions/%s/downloads", c.url, datasetID, edition, version)
-	clientlog.Do("updating version downloads", service, uri)
+// PutVersion update the version
+func (c *Client) PutVersion(datasetID string, edition string, version string, v Version) error {
```
nshumoogum: You can reduce the length of line by defining the type `string` once, such as:

`func (c *Client) PutVersion(datasetID, edition, version string, v Version) error {`

PR #52, opened 2017-12-14
Feature/healthcheck

CarlHembrough:
* Extract common code for calling health checks on regular intervals (ticker.go)
* Extract common code for creating a HTTP server for health checks (server.go)
* Created a healthcheck.Client implementation for Neo4j
* Change the logging of the existing healthcheck code to log when the healthcheck is done rather than when the endpoint is called.
* Add a readme to document how to use the code in various scenarios


neo4j/healthcheck.go
```diff
@@ -0,0 +1,54 @@
...
+	defer conn.Close()
+
+	rows, err := conn.QueryNeo(pingStmt, nil)
+	if err != nil {
+		log.ErrorC("neo4j healthcheck query", err, logData)
+		return neo4j.serviceName, err
+	}
+	defer rows.Close()
+
+	_, _, err = rows.All()
```
Matt-Guest: Is this chunk need? As we already know if the connection is open or closed.

neo4j/healthcheck.go
```diff
@@ -0,0 +1,54 @@
...
+func NewHealthCheckClient(dbPool bolt.ClosableDriverPool) *HealthCheckClient {
+
+	return &HealthCheckClient{
+		dbPool:      dbPool,
+		serviceName: "neo4j",
+	}
+}
+
+// Healthcheck calls neo4j to check its health status.
+func (neo4j *HealthCheckClient) Healthcheck() (string, error) {
```
Matt-Guest: Instead of a string / error, could we follow gedge's struct? https://github.com/ONSdigital/dp-dataset-api/blob/feature/healthcheck/api/healthcheck.go#L19

healthcheck/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+Make sure you call close on shutdown:
+
+```
+err = healthChecker.Close(ctx)
+```
+
+#### Add a health check to a service with an existing HTTP server
+
+Register the heath check handler as a route:
```
gedge: typo `health`

healthcheck/healthcheck.go
```diff
@@ -42,6 +42,8 @@ func MonitorExternal(clients ...Client) {
 	for _, client := range clients {
 		go func(client Client) {
 			if name, err := client.Healthcheck(); err != nil {
+				err := fmt.Errorf("unsuccessful healthcheck for %s: %v", name, err)
```
gedge: I thought we deprecated use of `fmt.Errorf` (makes searches on kibana more tricky), how about:
```
log.ErrorC(""unsuccessful healthcheck", err, log.Data{"external_service": name})
```

healthcheck/README.md
```diff
@@ -0,0 +1,55 @@
...
+=============
+
+#### Add a health check to a service without a HTTP server
+
+Create a health check server to check neo4j and the filter API every 30 seconds:
+```
+neoHealthCheck := neo4j.NewHealthCheckClient(neo4jConnPool)
+filterAPIHealthChecker := filterAPIClient.New(config.FilterAPIURL)
+
+healthChecker := healthcheck.NewServer(config.BindAddr, time.Second * 30, errorChannel,
```
gedge: the time element was not clear to me, consider:
```
healthChecker := healthcheck.NewServer(
    config.BindAddr,
    config.HealthCheckInterval,
    errorChannel,
    filterAPIHealthChecker,
    neoHealthCheck,
)
```
(which happens to be my preferred formatting, too!)

neo4j/healthcheck.go
CarlHembrough: Good point - I have removed it.

neo4j/healthcheck.go
CarlHembrough: I have used the interface already used by the health checks in go-ns. We could look at improving this.

healthcheck/README.md
CarlHembrough: Fixed 👍

healthcheck/healthcheck.go
CarlHembrough: Done 👍

healthcheck/README.md
CarlHembrough: Done 👍

PR #54, opened 2017-12-20
healthcheck.Do returns 429 if tests not run yet

gedge:
this is to let consul know that we've not acheived a good state yet
so it is essentially a warning


healthcheck/healthcheck.go
```diff
@@ -91,6 +91,9 @@ func Do(w http.ResponseWriter, req *http.Request) {
 		for stateKey := range healthState {
 			*(healthStateInfo.Errors) = append(*(healthStateInfo.Errors), healthError{Namespace: stateKey, ErrorMessage: healthState[stateKey].Error()})
 		}
+	} else if healthLastChecked.IsZero() {
+		w.WriteHeader(http.StatusTooManyRequests)
```
LloydGriffiths: 🌮

PR #57, opened 2017-12-31
 Add async kafka consumer

CarlHembrough:
### What
AsyncConsumer provides a higher level of abstraction over a Kafka consumer. Instead of dealing with a channel of messages, the AsyncConsumer allows a function to be provided that is called for each message consumed. It maintains its own go routine to consume messages, and provides a simple graceful shutdown with a blocking close method.

A usage example can be seen here: https://github.com/ONSdigital/dp-import-api/blob/feature/hierarchy-builder/event/observations_imported_consumer.go

```
// Create ConsumerGroup as before
consumer, err := kafka.NewConsumerGroup(...)
	
asyncConsumer := kafka.NewAsyncConsumer()
	
// Pass the consumer and the handler function into the AsyncConsumer consume function
asyncConsumer.Consume(consumer, func(message kafka.Message) {
	// handle message here
})
	
// close the AsyncConsumer first
asyncConsumer.Close(context)
consumer.Close(context)
```

### How to review

Review code / give it a try

### Who can review

Anyone


gedge:
I'm hoping that this work is unnecessary and this PR can be ignored - the functionality is already in `go-ns/kafka` and I'm currently testing this on [this branch](https://github.com/ONSdigital/dp-import-api/pull/32)

nshumoogum:
This functionality has already been implemented, closing pr and deleting branch

PR #59, opened 2018-01-09
Add elasticsearch healthcheck

nshumoogum:
### What

Add elasticsearch healthcheck to abide by the current healthcheck interface within go-ns. Updated healthcheck README.md to include new healthcheck.

### How to review

Check logic and one can test the code against dp-search-builder on branch `feature/initial-service` whereby the code is already integrated. You will need to be running elasticsearch on port 9200 or set environment variable to match where you are running elasticsearch instance (see README.md for env var name).

### Who can review

Team B

### DIFF
--- a/elasticsearch/healthcheck.go
+++ b/elasticsearch/healthcheck.go
@@ -1,0 +1,113 @@
+package elasticsearch
+
+import (
+	"encoding/json"
+	"errors"
+	"io/ioutil"
+
+	"github.com/ONSdigital/go-ns/healthcheck"
+	"github.com/ONSdigital/go-ns/rhttp"
+
+	"net/http"
+	"net/url"
+
+	"github.com/ONSdigital/go-ns/log"
+)
+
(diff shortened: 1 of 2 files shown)


elasticsearch/healthcheck.go
```diff
@@ -0,0 +1,113 @@
...
+
+// HealthCheckClient provides a healthcheck.Client implementation for health checking elasticsearch.
+type HealthCheckClient struct {
+	cli         *rhttp.Client
+	path        string
+	serviceName string
+}
+
+// ClusterHealth represents the response from the elasticsearch cluster health check
+type ClusterHealth struct {
```
Matt-Guest: Do we need the other items in the struct? as this only checks for the status?

nshumoogum: I left all the items available from request just in case the reviewer came up with other things to check before saying elastic search is okay. I will now remove unless there are any suggestions?

PR #61, opened 2018-01-18
Create a search API client

mattrout92:
### What

Add a search api client 

- Add Dimension search method
- Add support for authenticated requests
- Test coverage
- Search API models
- Healthcheck method

### How to review

Verify client works based on search api spec

### Who can review

Anyone


clients/search/data.go
```diff
@@ -0,0 +1,32 @@
...
+type Model struct {
+	Count      int    `json:"count"`
+	Items      []Item `json:"items"`
+	Limit      int    `json:"limit"`
+	Offset     int    `json:"offset"`
+	TotalCount int    `json:"total_count"`
+}
+
+// Item represents a single hierarchy item returned by the search api
+type Item struct {
```
nshumoogum: I think the json tags for item should have omitempty in case no results are found, like so:

`Code               string  `json:"code,omitempty"`

The reason for not setting the model struct keys to omitempty is because there is value in those fields being set to 0, i.e. you can set a different message to the web user, no results found for search term

clients/search/search.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// Dimension allows the searching of a dimension for a specific dimension option, optionally
+// pass in offset and limit parameters respectively
+func (c *Client) Dimension(datasetID, edition, version, name, query string, params ...int) (m *Model, err error) {
+	offset := offset
+	limit := limit
+
+	if len(params) > 0 {
+		offset = params[0]
```
nshumoogum: Should the params be an array of key value pairs, what if the parameters passed into client come in the wrong order?

clients/search/search.go
```diff
@@ -0,0 +1,133 @@
...
+		return service, &ErrInvalidSearchAPIResponse{http.StatusOK, resp.StatusCode, c.url + "/healthcheck"}
+	}
+
+	return service, nil
+}
+
+// Dimension allows the searching of a dimension for a specific dimension option, optionally
+// pass in offset and limit parameters respectively
+func (c *Client) Dimension(datasetID, edition, version, name, query string, params ...int) (m *Model, err error) {
+	offset := offset
```
nshumoogum: This would be clearer in my eyes

```
const (
  defaultOffset = 0
  defaultLimit = 0
)

.....

offset := defaultOffset
limit := defaultLimit
```

clients/search/search.go
```diff
@@ -0,0 +1,133 @@
...
+	limit := limit
+
+	if len(params) > 0 {
+		offset = params[0]
+	}
+	if len(params) > 1 {
+		limit = params[1]
+	}
+
+	uri := fmt.Sprintf("%s/search/datasets/%s/editions/%s/versions/%s/dimensions/%s?q=%s&limit=%d&offset=%d",
```
nshumoogum: Probably need to sanitise uri by encoding it

PR #65, opened 2018-01-26
[MS] Added a default implementation of Client to healthcheck

YetAnotherMatt:
Added a default implementation of Client to healthcheck


healthcheck/client.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/ONSdigital/go-ns/rhttp"
+	"net/http"
+	"fmt"
+)
+
+// healthcheckClient is an implementation of Client that can be used to call the healthcheck endpoint of any service
+type healthcheckClient struct {
+	Client  httpClient
+	Url     string
+	Service string
```
CarlHembrough: Consider making these struct members private to encapsulate them. When they are private they must be injected via the constructor function

healthcheck/client.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// httpClient has a get method. Implemented by http.Client etc.
+type httpClient interface {
+	Get(url string) (*http.Response, error)
+}
```
CarlHembrough: It wasn't feeling right that this interface would be private - it's because I think it should be a parameter in the constructor function. For that reason I would make it public and have the mock generation directive in this file for completeness. (sorry I did not spot this earlier)

healthcheck/client.go
```diff
@@ -0,0 +1,61 @@
...
+type errorResponse struct {
+	service      string
+	expectedCode int
+	actualCode   int
+	uri          string
+}
+
+// NewClient creates a new Client for a service with the given name and healthcheck endpoint (url).
+// it uses rhttp.Client to make requests to the given url, reporting on any response status code != 200.
+func NewClient(service string, url string) *healthcheckClient {
```
CarlHembrough: Consider adding a parameter for the http client, which will be the HttpClient interface type

healthcheck/client_test.go
```diff
@@ -0,0 +1,59 @@
+package healthcheck
```
CarlHembrough: This one is a matter of preference as I know not everyone agrees with me. However, I like to have my tests in a different package from the code I am testing. This forces you to test via only the public interface of the package and not poke around internally. i.e. consider using the constructor function to instantiate the healthcheck client instead of the struct initialiser. (you will have to inject the http client into the constructor also)

healthcheck/client_test.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"errors"
+	"net/http"
+	"testing"
+
+	"github.com/ONSdigital/go-ns/healthcheck/mock_healthcheck"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestUnitHealthcheckClient(t *testing.T) {
```
CarlHembrough: Seems a bit odd to have 'Unit' in the name here. I see its used elsewhere in the package though. If using golang IDE you can press `CTRL + space` after the keyword `func` and it will suggest test names in the typical convention.

healthcheck/client_test.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		client := healthcheckClient{Client:mock, Url:"http://foo/bar", Service: service}
+
+		s, e := client.Healthcheck()
+		So(s, ShouldEqual, service)
+		So(e, ShouldBeNil)
+	})
+
+	Convey("Healthcheck returns error when status != 200", t, func() {
```
CarlHembrough: This may also be a matter of preference but I find it easier to work with the tests if you put each top level `Convey` in a separate function. When you are debugging tests its much nicer being able to run them isolated.

healthcheck/client_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/ONSdigital/go-ns/healthcheck/mock_healthcheck"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestUnitHealthcheckClient(t *testing.T) {
+
+	service := "myService"
+
+	Convey("Healthcheck returns nil when no error", t, func() {
```
CarlHembrough: Consider using a given, when, then format when writing tests. It seems verbose at first, but makes for a nice consistent format - see an example here: https://github.com/ONSdigital/dp-dataset-exporter/blob/04a99a7d6fa0999f9e01ca815d707c2e1884df2a/filter/store_test.go#L32

healthcheck/client_test.go
```diff
@@ -0,0 +1,59 @@
...
+		mock := &mock_healthcheck.HttpClientMock{
+			GetFunc: func(url string) (*http.Response, error) {
+				return &http.Response{StatusCode:http.StatusOK}, nil
+			},
+		}
+
+		client := healthcheckClient{Client:mock, Url:"http://foo/bar", Service: service}
+
+		s, e := client.Healthcheck()
```
CarlHembrough: It is certainly conventional to have short variable names for locally scoped variables, however I would consider using `err` for error variables. This seems to be widely used.

healthcheck/client_test.go
```diff
@@ -0,0 +1,59 @@
...
+		mock := &mock_healthcheck.HttpClientMock{
+			GetFunc: func(url string) (*http.Response, error) {
+				return &http.Response{StatusCode:http.StatusInternalServerError}, err
+			},
+		}
+
+		client := healthcheckClient{Client:mock, Url:"http://foo/bar", Service: service}
+
+		s, e := client.Healthcheck()
+		So(s, ShouldEqual, service)
```
CarlHembrough: The generated mocks also maintain a list of calls to each function on the mock. Its quite common to assert the number of calls to a mock function using: `So(len(mock.GetCalls()), ShouldEqual, 1)`

PR #68, opened 2018-03-05
Feature/vault

Matt-Guest:
### What

Created new Vault client library in go-ns. Includes some documentation and examples using and generating tokens.

### How to review

Read the documentation, review the code, run the code.

### Who can review

anyone


vault/client.go
```diff
@@ -0,0 +1,77 @@
...
+	config := vaultapi.Config{Address: vaultAddress, MaxRetries: retries}
+	client, err := vaultapi.NewClient(&config)
+	if err != nil {
+		return nil, err
+	}
+	client.SetToken(token)
+	return &VaultClient{token, client}, nil
+}
+
+// CreateTLSVaultClient by providing a auth token, vault address and the maxium number of retries for a request
```
gedge: this name irks me - does `CreateVaultClientTLS` sound better to you?

vault/client.go
```diff
@@ -0,0 +1,77 @@
...
+	vaultapi "github.com/hashicorp/vault/api"
+)
+
+// VaultClient Used to read and write secrets from vault
+type VaultClient struct {
+	token  string
+	client *vaultapi.Client
+}
+
+// CreateVaultClient by providing a auth token, vault address and the maxium number of retries for a request
```
gedge: typos `a auth token` and `maxium`?

vault/client.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"errors"
+
+	vaultapi "github.com/hashicorp/vault/api"
+)
+
+// VaultClient Used to read and write secrets from vault
+type VaultClient struct {
+	token  string
```
gedge: is `token` ever used (e.g. by a getter)?

vault/client.go
gedge: same typos as above

vault/client.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	client.SetToken(token)
+	return &VaultClient{token, client}, nil
+}
+
+// Read a secret from vault. If the token does not have the correct policy this shall return an error or if the
+// vault server is not reachable. This method shall return all the information store about the secret
```
gedge: did you mean
```
// Read reads a secret from vault. If the token does not have the correct policy this returns an error;
// if the vault server is not reachable, return all the information stored about the secret.
```

vault/client.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	val, ok := data[key]
+	if !ok {
+		return "", errors.New("key not found")
+	}
+	return val.(string), nil
+}
+
+// Write a secret to vault. If the token does not have the correct policy this shall return an error or if the
+// vault server is not reachable. If nil is returned the operation was successful
```
gedge:
```
// Write writes a secret to vault. Returns an error if the token does not have the correct policy or the
// vault server is not reachable. Returns nil when the operation was successful.
```

vault/example/Makefile
```diff
@@ -0,0 +1,22 @@
+
+VAULT_ADDR='http://127.0.0.1:8200'
```
gedge: make this conditional (defaulted to) with `VAULT_ADDR?='http://127.0.0.1:8200'`

vault/example/example.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/ONSdigital/go-ns/log"
+	"github.com/ONSdigital/go-ns/vault"
+)
+
+func main() {
+
+	log.Namespace = "vault-example"
+	log.Debug("Creating vault client", log.Data{})
+	devAddress := "http://127.0.0.1:8200"
```
gedge: this should be `vaultAddress := os.Getenv("VAULT_ADDR")`

vault/client.go
Matt-Guest: I renamed the function and fixed the typo

vault/client.go
Matt-Guest: done

vault/client.go
Matt-Guest: No its never used. I removed it 👍

vault/client.go
Matt-Guest: Thanks

vault/client.go
Matt-Guest: 👍

vault/example/Makefile
Matt-Guest: done

vault/example/example.go
Matt-Guest: done

Matt-Guest:
Is been vendored already? Is there anything missing you can see?

PR #72, opened 2018-03-12
Feature/dataset client with context

mattrout92:
### What

- Add rchttp to dataset client
- Add vault healthcheck
- Add download specific authentication
- Updates to the edition model to account for `next` when authenticated


clients/dataset/dataset.go
```diff
@@ -84,7 +101,8 @@ func (c *Client) SetInternalToken(token string) {
 
 func (c *Client) setInternalTokenHeader(req *http.Request, cfg ...Config) {
```
eldeal: Should this be renamed to more generic 'setHeaders' or somethign?

PR #73, opened 2018-03-15
Feature/identity middleware

CarlHembrough:
### What

Update identity middleware to call new identity endpoint and store response in request context.

* Added tests around existing functionality
* Read `Authorization` header from requests as well as existing Florence token
* If both headers are provided then the user header takes precedence
* Store `User-Identity` and `Caller-Identity` in request context to use in downstream calls.

### How to review

Review changes

### Who can review

Anyone


identity/identity.go
```diff
@@ -6,49 +6,110 @@ import (
...
 
-			h.ServeHTTP(w, req)
+			h.ServeHTTP(w, req.WithContext(ctx))
 		})
 	}
 }
+
+func unmarshalResponse(resp *http.Response) (*identityResponse, error) {
+
+	var identityResp *identityResponse
```
mattrout92: consider naming the variables in the response part of function signature, e.g:

```go
func unmarshalResponse(resp *http.Response) (identityResp *identityResponse, err error) {
```

identity/identity.go
```diff
@@ -6,49 +6,110 @@ import (
...
 
+					resp, err := cli.Do(ctx, zebReq)
+					if err != nil {
+						log.ErrorR(req, err, nil)
+						w.WriteHeader(http.StatusInternalServerError)
+						return
+					}
+
+					// Check to see if the florence user is authorised
+					if resp.StatusCode != http.StatusOK {
```
mattrout92: May want to log that an unexpected status code was returned

identity/identity.go
```diff
@@ -6,49 +6,110 @@ import (
...
+	"context"
 	"github.com/ONSdigital/go-ns/log"
+	"encoding/json"
+	"io/ioutil"
 )
 
+//go:generate moq -out identitytest/http_client.go -pkg identitytest . HttpClient
+
 var zebedeeURL = os.Getenv("ZEBEDEE_URL")
+var florenceHeaderKey = "X-Florence-Token"
```
mattrout92: apart from zebedeeURL I think these are constants

identity/identity.go
```diff
@@ -6,49 +6,110 @@ import (
...
+					zebReq, err := http.NewRequest("GET", zebedeeURL+"/identity", nil)
+					if err != nil {
+						log.ErrorR(req, err, nil)
+						w.WriteHeader(http.StatusInternalServerError)
+						return
+					}
+
+					if isUserReq {
+						zebReq.Header.Set(florenceHeaderKey, florenceToken)
+					} else if isServiceReq {
```
mattrout92: don't think you need the second if here, else will do

identity/identity.go
CarlHembrough: Done 👍

identity/identity.go
CarlHembrough: Done

identity/identity.go
CarlHembrough: Changed to const

identity/identity.go
CarlHembrough: Good point - done

PR #74, opened 2018-03-16
Add authentication wrapper function and readme

CarlHembrough:
### What

All existing API's are using the same function to wrap endpoints with an auth check. I have updated the function to work with identities stored in request context and added it to go-ns for reuse.

Added a readme to document the usage of the middleware and wrapper function.

### How to review

Review changes

### Who can review

Anyone

### DIFF
--- a/identity/authentication.go
+++ b/identity/authentication.go
@@ -11,14 +11,19 @@ import (
 func Check(handle func(http.ResponseWriter, *http.Request)) http.HandlerFunc {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 
+		log.DebugR(r, "checking for an identity in request context", nil)
+
 		callerIdentity := r.Context().Value(callerIdentityKey)
+		logData := log.Data{ "caller_identity": callerIdentity }
 
 		// just checking if an identity exists until permissions are being provided.
 		if callerIdentity == nil || callerIdentity == ""{
 			http.Error(w, "requested resource not found", http.StatusNotFound)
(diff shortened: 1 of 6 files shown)


identity/authentication.go
```diff
@@ -0,0 +1,26 @@
...
+		callerIdentity := r.Context().Value(callerIdentityKey)
+
+		// just checking if an identity exists until permissions are being provided.
+		if callerIdentity == nil || callerIdentity == ""{
+			http.Error(w, "requested resource not found", http.StatusNotFound)
+			log.Error(errors.New("client missing auth token in header"), nil)
+			return
+		}
+
+		// The request has been authenticated, now run the clients request
```
daiLlew: Is it worth logging that the request has been authenticated?

CarlHembrough: Added a log statement at the start of the handler and at the end when its complete

identity/authentication_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+	Convey("Given a context with an identity", t, func() {
+
+		ctx := context.WithValue(context.Background(), callerIdentityKey, "user@ons.gov.uk")
+
+		Convey("When IsPresent is called with the context", func() {
+
+			identityIsPresent := IsPresent(ctx)
+
+			Convey("Then a 404 response is returned", func() {
```
gedge: incorrect statement or test

CarlHembrough: Copy paste fail - I have updated the statement 👍

PR #81, opened 2018-04-16
Feature/auth check separate client

gedge:
### What

- `clients/...` changes:
  - `clients/identity` is new - does *not* add own service-token, but uses incoming request context
  - `clients/dataset` fully moved to new rchttp (using context for auth)
- `healthcheck/client` updated to use `rchttp`

- `common/...` changes:
  - `common/identity` - moved a lot of common strings/funcs to here
  - `common/rchttp` - the  interface for which is shared here

- `identity/...` - client moved to `clients/identity` (see above)

- `rchttp` changes:
  - service token (now a field/property of rchttp) added automatically to request headers
  - "Bearer " added automatically to service token when put into header
  - user-id (from context) added automatically to headers
  - `cli := rchttp.NewClient()` now preferred over `cli := rchttp.DefaultClient` (which was shared)
    - or better `cli := rchttp.ClientWithServiceToken(serviceToken)` if not from incoming request
  - mocking also shared by `common/mock_common` (mockgen) and `common/commontest` (moq)
  - lots more testing

### How to review

Check and run the tests.

### Who can review

reviewer.user != "gedge" && reviewer.group == "cmd"


clients/dataset/dataset.go
```diff
@@ -83,60 +43,42 @@ var _ error = ErrInvalidDatasetAPIResponse{}
...
-	url           string
-	internalToken string
+	cli common.RCHTTPClienter
+	url string
 }
 
-// New creates a new instance of Client with a given dataset api url
-func New(datasetAPIURL string) *Client {
+// NewAPIClient creates a new instance of Client with a given dataset api url and the relevant tokens
+func NewAPIClient(datasetAPIURL, serviceToken, xDownloadServiceToken, florenceToken string) *Client {
```
CarlHembrough: It seems odd to have a Florence token injected into the constructor - won't this potentially change for each request if there is a different user? I would imply that a new client is needed for each request?

CarlHembrough: I am not familiar with the xDownloadServiceToken - is this something other than the service token?

gedge: you're right, I think I got carried away with fields to put into the client, I'll take it out

gedge: yes, it's essentially a token specific to the download service (why it's not using the usual `Authentication:` header pattern, I don't recall: Matt will know)

rchttp/client.go
```diff
@@ -10,17 +11,21 @@ import (
...
-	ExponentialBackoff bool
-	RetryTime          time.Duration
-	HTTPClient         *http.Client
+	MaxRetries           int
+	ExponentialBackoff   bool
+	RetryTime            time.Duration
+	HTTPClient           *http.Client
+	AuthToken            string
+	DownloadServiceToken string
+	FlorenceToken        string
```
CarlHembrough: I can't see why the Florence Token /  DownloadServiceToken are stored within the client here. They don't appear to be used within rchttp.client. It appears that  the florence token is injected into the client as required for requests, so I think we can remove it from the struct / remove the setters?

clients/dataset/dataset.go
```diff
@@ -83,60 +43,42 @@ var _ error = ErrInvalidDatasetAPIResponse{}
...
+			xDownloadServiceToken,
+		),
+		url: datasetAPIURL,
 	}
 }
 
 // Healthcheck calls the healthcheck endpoint on the api and alerts the caller of any errors
 func (c *Client) Healthcheck() (string, error) {
-	resp, err := c.cli.Get(c.url + "/healthcheck")
+	ctx := context.Background()
```
CarlHembrough: Do we want to allow context to be passed into the healthcheck for cancellation? (possibly a future enhancement rather than something to add right now)

common/client.go
```diff
@@ -0,0 +1,14 @@
...
+import "sync"
+
+type APIClient struct {
+	BaseURL    string
+	AuthToken  string
+	HTTPClient RCHTTPClienter
+	Lock       sync.RWMutex
+}
+
+type APIClienter interface {
```
CarlHembrough: I can't see that this is used anywhere?

common/rchttp.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+//go:generate mockgen -destination mock_common/rchttp_client.go github.com/ONSdigital/go-ns/common RCHTTPClienter
+//go:generate moq -out commontest/rchttp_client.go -pkg commontest . RCHTTPClienter
+
+// RCHTTPClienter provides an interface for methods on an HTTP Client
+type RCHTTPClienter interface {
+	SetAuthToken(token string)
+	SetDownloadServiceToken(token string)
+	SetFlorenceToken(token string)
```
CarlHembrough: Remove?

rchttp/client.go
gedge: As per our discussion, I will remove FlorenceToken (it's set by `clients/identity` not rchttp)

common/client.go
gedge: Agreed.

common/identity.go
```diff
@@ -0,0 +1,98 @@
...
+
+// AddDeprecatedHeader sets the deprecated header on the given request
+func AddDeprecatedHeader(r *http.Request, token string) {
+	if len(token) > 0 {
+		r.Header.Add(DeprecatedAuthHeader, token)
+	}
+}
+
+// IsPresent determines if an identity is present on the given context.
+func IsPresent(ctx context.Context) bool {
```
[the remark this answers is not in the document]
gedge: Totally agree. I prefer `IsCallerPresent(...)` ?

rchttp/client.go
```diff
@@ -43,15 +48,78 @@ var DefaultClient = &Client{
...
+}
 
 // Do calls ctxhttp.Do with the addition of exponential backoff
 func (c *Client) Do(ctx context.Context, req *http.Request) (*http.Response, error) {
+
+	if len(c.AuthToken) > 0 {
+		// only add this header if not already set (e.g. for authClient)
+		if len(req.Header.Get(common.AuthHeaderKey)) == 0 {
+			common.AddServiceTokenHeader(req, c.AuthToken)
+		}
```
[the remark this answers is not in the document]
gedge: Agreed.

clients/dataset/dataset.go
gedge: I am leaning towards: _future!_

common/identity.go
```diff
@@ -50,6 +50,13 @@ func AddServiceTokenHeader(r *http.Request, serviceToken string) {
 	}
 }
 
+// AddDownloadServiceTokenHeader sets the given download service token on the given request
+func AddDownloadServiceTokenHeader(r *http.Request, serviceToken string) {
+	if len(serviceToken) > 0 {
+		r.Header.Add(DownloadServiceHeaderKey, BearerPrefix+serviceToken)
```
CarlHembrough: No bearer required here 🐻

gedge: there is nothing sitting in the woods

PR #82, opened 2018-04-17
Feature/update avro unmarshal

nshumoogum:
### What

Add functionality to the unmarshal method to handle objects nested within an object

### How to review

Make sure all unit tests pass and that the code makes logical sense.

Note: The commented out lines in the unit tests are placeholders for missing tests, these will be written when we have time to do so.

### Who can review

Team B

### DIFF
--- a/avro/test_data.go
+++ b/avro/test_data.go
@@ -69,7 +69,8 @@ var nestedObjectSchema = `{
 				"fields": [
 					{
 			  		"name": "surname",
-						"type": "string"
+						"type": "string",
+            "default": ""
 			 		},
 					{
 					  "name": "middle-name",
@@ -78,7 +79,8 @@ var nestedObjectSchema = `{
 				  },
 					{
 						"name": "forename",
-						"type": "string"
+						"type": "string",
+            "default": ""
 				  }
 				]
 			}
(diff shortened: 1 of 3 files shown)


avro/test_data.go
```diff
@@ -53,6 +53,61 @@ var testNestedArraySchema = `{
...
+		{
+		  "name": "footballer",
+			"type": {
+			  "name": "footballer-name",
+			  "type": "record",
+				"fields": [
+					{
+			  		"name": "surname",
+						"type": "string",
+            "default": ""
```
mattrout92: spacing

avro/test_data.go
```diff
@@ -53,6 +53,61 @@ var testNestedArraySchema = `{
...
+			 		},
+					{
+					  "name": "middle-name",
+						"type": "string",
+						"default": ""
+				  },
+					{
+						"name": "forename",
+						"type": "string",
+            "default": ""
```
mattrout92: spacing

avro/test_data.go
nshumoogum: I saw this and I swear I used go fmt........ :(

nshumoogum: Just realised its in a string so it won't format it, I have to do it manually :(

PR #83, opened 2018-04-17
Feature/audit

daiLlew:
### What
Initial attempt at audit package.


audit/audit.go
```diff
@@ -0,0 +1,130 @@
...
+				ctx = context.WithValue(ctx, fromContextKey, from)
+				event.User = from
+			}
+
+			event.RequestID = r.Header.Get(requestIDHeader)
+			event.Namespace = a.Namespace
+
+			eventString, err := json.Marshal(event)
+			if err != nil {
+				//we couldn't create the audit event. some input must be invalid
```
nshumoogum: This shouldn't happen if we set json tags against the event struct, where the field can be empty set `omitempty` (see below), also I would agree if there is an error here we would want to logout the error and force an internal server error in the response, `w http.ResponseWriter`

```
type Event struct {
    Created          string                `avro:"created"            json:"created,omitempty"`
    Service          string                `avro:"service"            json:"service"`
    Namespace        string                `avro:"name_space"         json:"name_space"`
    RequestID        string                `avro:"request_id"         json:"request_id"`
    User             string                `avro:"user"               json:"user"`
    AttemptedAction  string                `avro:"attempted_action"   json:"attempted_action,omitempty"`
    Result           string                `avro:"result"             json:"result,omitempty"`
    Params           []keyValuePair        `avro:"params"             json:"params"`
}

type keyValuePair struct {
    Key    string `avro:"key"      json:"key,omitempty"`
    Value  string `avro:"value"    json:"value,omitempty"`
}
```

audit/audit.go
```diff
@@ -0,0 +1,130 @@
...
+
+			if from := r.Header.Get(fromHeader); len(from) > 0 {
+				ctx = context.WithValue(ctx, fromContextKey, from)
+				event.User = from
+			}
+
+			event.RequestID = r.Header.Get(requestIDHeader)
+			event.Namespace = a.Namespace
+
+			eventString, err := json.Marshal(event)
```
nshumoogum: `eventString` is []byte not string, so it would be better to call is either just eventBytes

It will read better when you stringify the byte array like so `string(eventBytes)` on line 84

audit/audit.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+type contextKey string
+
+const eventContextKey = contextKey("audit")
+const fromContextKey = contextKey("from")
+const requestIDHeader = "X-Request-Id"
+const fromHeader = "from"
+
+type Event struct {
```
nshumoogum: These should have json tags, see comment below

audit/audit.go
```diff
@@ -0,0 +1,130 @@
...
+	log.Info("logging audit message...", nil)
+	out := a.Producer.Output()
+	out <- eventBytes
+}
+
+func GetEvent(input context.Context) *Event {
+	a := &Event{}
+	if s, ok := input.Value(eventContextKey).(string); ok {
+		if err := json.Unmarshal([]byte(s), a); err != nil {
+			log.ErrorC("error while attempting to unmarshal audit event from context", err, nil)
```
nshumoogum: This should return error

audit/audit.go
```diff
@@ -0,0 +1,164 @@
...
+
+//go:generate moq -out generated_mocks.go -pkg audit . AuditorService OutboundProducer
+
+type contextKey string
+
+const (
+	eventContextKey = contextKey("audit")
+	fromContextKey  = contextKey("from")
+	requestIDHeader = "X-Request-Id"
+	fromHeader      = "from"
```
CarlHembrough: Not entirely sure what the from header is for. It looks like it will be the user who made the request? If so, there is already a `User-Identity` header for this. See here for an example of it being read https://github.com/ONSdigital/go-ns/blob/4001bc2311e9cd2de98a365b6ff5ca6b1065e576/clients/identity/authentication.go#L91

CarlHembrough:
Need to merge in latest changes from master

audit/audit.go
```diff
@@ -0,0 +1,164 @@
...
+	Value string `avro:"value" json:"value,omitempty"`
+}
+
+// OutboundProducer defines a producer for sending outbound audit events to a kafka topic
+type OutboundProducer interface {
+	Output() chan []byte
+}
+
+//AuditorService defines the behaviour of an audior
+type AuditorService interface {
```
CarlHembrough: As your providing an interface, I find its useful to provide a compile time check that the Auditor implements the interface:
```
// ensure that Auditor satisfies the AuditorService interface.
var _ AuditorService = (*Auditor)(nil)
```

audit/audit.go
```diff
@@ -0,0 +1,164 @@
...
+	GetEvent(input context.Context) (*Event, error)
+	Record(ctx context.Context, action string, result string, params Params) error
+}
+
+//Auditor provides functions for interception HTTP requests and populating the context with a base audit event and
+// recording audit events
+type Auditor struct {
+	namespace     string
+	tokenName     string
```
CarlHembrough: This looks like its the header name for service tokens - which is always `Authorization`. Consider removing tokenName from the struct and using the const defined here: https://github.com/ONSdigital/go-ns/blob/4001bc2311e9cd2de98a365b6ff5ca6b1065e576/common/identity.go#L14

audit/audit.go
```diff
@@ -0,0 +1,164 @@
...
+	e.AttemptedAction = action
+	e.Result = result
+
+	if params != nil {
+		for k, v := range params {
+			e.Params = append(e.Params, keyValuePair{Key: k, Value: v})
+		}
+	}
+
+	e.Created = time.Now().String()
```
gedge: This will create a string which includes any current TZ offset (e.g. BST on dev machines, but likely UTC on the servers) - do we want to enforce UTC for all logging for consistency?

I think UTC times are more commonly used in other logging.

Perhaps:  `e.Created = time.Now().UTC().Format(time.RFC3339Nano)`

audit/audit_dummy.go
```diff
@@ -0,0 +1,40 @@
+package audit
+
+import (
+	"encoding/json"
+	"github.com/ONSdigital/go-ns/log"
+)
+
+type DummyProducer struct {
```
gedge: where is this file used? has this file been superceded by OutboundProducerMock?

audit/audit_test.go
```diff
@@ -0,0 +1,202 @@
...
+func TestAuditor_Record(t *testing.T) {
+	Convey("given no audit event exists in the provided context", t, func() {
+		producer := &OutboundProducerMock{}
+
+		auditor := New(producer, namespace, "test")
+
+		// record the audit event
+		err := auditor.Record(context.Background(), auditAction, auditResult, nil)
+
+		expectedErr := auditError("no event found in context.Context", auditAction, nil)
```
gedge: I can't help but feel that this test is too dependent on `auditError` working, and if it breaks (future change), then these tests will be testing the broken output against the broken output - and hence probably succeed.

Perhaps add a new `func Test_auditError(...)` before all these to test that `auditError` returns something sane/expected/non-"".

audit/audit_test.go
```diff
@@ -0,0 +1,202 @@
...
+		err := auditor.Record(nil, auditAction, "", nil)
+
+		So(len(producer.OutputCalls()), ShouldEqual, 0)
+		expectedErr := auditError("result is required field", "test", nil)
+		So(err.Error(), ShouldEqual, expectedErr.Error())
+	})
+}
+
+func TestAuditor_GetEvent(t *testing.T) {
+	Convey("given a no event exists in the provided context", t, func() {
```
gedge: "given a " should be just "given" in most of these...

identity/identity.go
```diff
@@ -59,7 +59,6 @@ func HandlerForHTTPClient(doAuth bool, cli HTTPClient, zebedeeURL string) func(h
 				log.DebugR(req, "authentication enabled, checking for expected tokens", logData)
 
 				if isUserReq || isServiceReq {
-
```
gedge: this change will only make it worse when merging with `master` later (it causes the only conflict, I discovered)

please reinstate that line 👾

audit/audit_dummy.go
daiLlew: Didn't know we had one - this is just temp and will be delete shorty.

audit/audit_test.go
daiLlew: Good idea. I have to say I am not a fan of Go approach errors - sometimes you want to pass back values that give the error some context but the only way to do that is building up a string which feels a bit naff. I guess I could create an audit error type which takes these values but ultimately it is string concatenation again.

audit/audit.go
```diff
@@ -0,0 +1,164 @@
...
+	return &Auditor{
+		producer:      producer,
+		namespace:     namespace,
+		tokenName:     token,
+		marshalToAvro: EventSchema.Marshal,
+	}
+}
+
+//Interceptor returns an http.Handler which populates a http.Request.Context with a base audit event - fields common
+// to all events regardless of context before passing it on to the handler in the chain.
```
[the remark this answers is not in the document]
daiLlew: This should happen before as the identity handler will need to audit things (this is my understading) and will need the event to exist in the context. I will add the comment though

audit/audit.go
```diff
@@ -0,0 +1,164 @@
...
+func (a *Auditor) Interceptor() func(handler http.Handler) http.Handler {
+	return func(handler http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctx := r.Context()
+			event := Event{
+				RequestID: r.Header.Get(requestIDHeader),
```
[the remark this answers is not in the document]
daiLlew: Hmmm, we seem to heading down a road of a specific order for the middleware which I think is broader than just this change. It should like the reqID handler should always be the first in the chain followed by audit.

identity/identity.go
daiLlew: lol who knew whitespace could cause that much damage.

audit/schema.go
```diff
@@ -0,0 +1,53 @@
...
+      "type": "string",
+      "default": ""
+    },
+    {
+      "name": "action_result",
+      "type": "string",
+      "default": ""
+    },
+    {
+      "name": "params",
```
nshumoogum: So the default value for a field of type map should be null, so `params` should look something like:

```
{
  "name": "params",
  "default": null,
  "type": [
    "null",
    {
      "type": "map",
      "values": "string"
    }
  ]
}
```

See subordinate field [here](http://bytepadding.com/big-data/spark/avro/define-avro-schema-with-default-values/)

avro/avro.go
```diff
@@ -461,3 +510,16 @@ func unmarshalStruct(val interface{}, v reflect.Value, i int) (interface{}, erro
 
 	return item.Interface(), nil
 }
+
+func unmarshalMap(value interface{}) map[string]string {
+	return unmarshalStringMap(value)
```
nshumoogum: `unmarshalMap` should handle value of nil and return nil, like so

```
if value == nil {
  return nil
}
```

nshumoogum: Extend unit tests for empty map

audit/schema.go
gedge: I'll look at this

avro/avro.go
gedge: I'll look at this

audit/audit.go
```diff
@@ -0,0 +1,179 @@
...
+
+	log.Info("logging audit message", log.Data{"auditEvent": e})
+	a.producer.Output() <- avroBytes
+	return nil
+}
+
+//NewAuditError creates new audit.Error with default field values where necessary and orders the params alphabetically.
+func NewAuditError(cause string, attemptedAction string, actionResult string, params common.Params) Error {
+	if cause == "" {
+		cause = nilStr
```
eldeal: Why do the default values need to be declared as "nil" rather than just left blank? What would the harm of just calling `Error{"this is a cause", "actionname", "", params}` rather than using the function to create the error be? Should this be documented for the function?

audit/audit.go
```diff
@@ -0,0 +1,179 @@
...
+
+	e.RequestID = requestID.Get(ctx)
+
+	avroBytes, err := a.marshalToAvro(e)
+	if err != nil {
+		log.Error(err, nil)
+		return NewAuditError("error marshalling event to arvo", attemptedAction, actionResult, params)
+	}
+
+	log.Info("logging audit message", log.Data{"auditEvent": e})
```
eldeal: Is there value in this log message being more explicit about this happening before we've sent the kafka message? just thinking when debugging it might help to be clear whether this log message means 'successful audit all the things' or 'weve created something in an audit structure, now we're going to try actually sending it'

daiLlew: TBH I only put this log message here so I could see that was something was being "audited" while I was developing it. I will remove it.

audit/audit.go
daiLlew: Fair point I will update to ""

CarlHembrough:
The latest changes from go-ns master branch need merging in - hopefully not too painful!

PR #85, opened 2018-04-20
Logout partial tokens in middleware

nshumoogum:
### What

Logout a part of tokens (albeit florence or authorization tokens) to identify where the request is coming from and that the service has the right tokens.

### How to review

Check logic makes sense

### Who can review

Team B


clients/identity/authentication_test.go
```diff
@@ -261,6 +261,22 @@ func TestHandler_bothTokens(t *testing.T) {
 	})
 }
 
+func TestSplitTokens(t *testing.T) {
+	Convey("Given a florence token and service token", t, func() {
+		florenceToken := "Bearer 987654321"
```
CarlHembrough: Not that it matters for the test - but to prevent confusion it may be worth swapping the values so the service token is prefixed with Bearer, instead of the florence token.

clients/identity/authentication.go
```diff
@@ -5,12 +5,21 @@ import (
...
 	"io/ioutil"
 	"net/http"
+	"strings"
 
 	"github.com/ONSdigital/go-ns/common"
 	"github.com/ONSdigital/go-ns/log"
 	"github.com/ONSdigital/go-ns/rchttp"
 )
 
+const prefix = "Bearer "
```
CarlHembrough: You don't have to define this - it's already defined here: https://github.com/ONSdigital/go-ns/blob/4001bc2311e9cd2de98a365b6ff5ca6b1065e576/common/identity.go#L19

clients/identity/authentication.go
```diff
@@ -101,6 +114,33 @@ func (api IdentityClient) CheckRequest(req *http.Request) (context.Context, int,
...
+	var ft, at tokenObject
+	if len(florenceToken) > 0 {
+		ft = splitToken(florenceToken)
+	}
+
+	if len(authToken) > 0 {
+		at = splitToken(authToken)
+	}
+
+	return log.Data{"florence_token": ft, "auth_token": at}
```
CarlHembrough: I wonder if its worth only adding to the log.Data if the token is not empty, as only one of the two tokens will be present?

clients/identity/authentication.go
```diff
@@ -101,6 +112,37 @@ func (api IdentityClient) CheckRequest(req *http.Request) (context.Context, int,
...
+
+func splitToken(token string) (tokenObj tokenObject) {
+
+	splitToken := strings.Split(token, " ")
+	tokenObj.numberOfParts = len(splitToken)
+	tokenObj.hasPrefix = strings.HasPrefix(token, common.BearerPrefix)
+	if len(token) > 6 {
+		tokenObj.tokenPart = token[len(token)-6:]
+	} else {
+		tokenObj.tokenPart = token
```
eldeal: am i reading this right that where we have a short token we would now be logging the whole thing?  i cant think of any cases where we'd have such short tokens but if we did i dont think its ever gonna be a good idea to log a full token 🤔

clients/identity/authentication.go
```diff
@@ -27,24 +34,28 @@ func NewAPIClient(cli common.RCHTTPClienter, url string) (api *IdentityClient) {
...
 	ctx := req.Context()
 
 	florenceToken := req.Header.Get(common.FlorenceHeaderKey)
 	authToken := req.Header.Get(common.AuthHeaderKey)
+	logData = splitTokens(florenceToken, authToken)
 
 	isUserReq := len(florenceToken) > 0
 	isServiceReq := len(authToken) > 0
+	logData["is_user_request"] = isUserReq
+	logData["is_service_request"] = isServiceReq
```
eldeal: it feels like this logic could be tidied up, as the two request types are mutually exclusive. the `if both nil` check could be moved up to before line 44, we dont need to split the tokens if they're nil, and the check can be done based on length.

gedge: I've moved the `logData` to when it's needed.

PR #86, opened 2018-04-26
Feature/audit avro map

gedge:
### What

use `map` params instead of (sorted) array for audit struct

### DIFF
--- a/audit/audit_test.go
+++ b/audit/audit_test.go
@@ -3,14 +3,16 @@ package audit
 import (
 	"context"
 	"fmt"
+	"strings"
+	"testing"
+	"time"
+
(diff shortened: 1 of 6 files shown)


audit/audit_test.go
```diff
@@ -234,3 +239,21 @@ func setUpContext() context.Context {
 	ctx = identity.SetUser(ctx, user)
 	return ctx
 }
+
+func extractMapFromString(str string) (params common.Params, err error) {
```
daiLlew: I don't like this. The caller shouldn't know about the internal details of the audit error which is why I added this logic to the `Error()` method. The caller then just calls the method and gets the (same) string back each time and doesn't need to know its an Audit error.

PR #89, opened 2018-05-01
Add basic middleware implementation for healthchecks.

CarlHembrough:
### What

Add basic middleware implementation for healthchecks. This will allow healthchecks to be injected before identity middleware

### How to review

Review code


healthcheck/middleware.go
```diff
@@ -0,0 +1,20 @@
+package healthcheck
+
+import (
+	"net/http"
+)
+
+// HandlerForHTTPClient allows a handler to be created that uses the given HTTP client
```
eldeal: comment doesnt match function

CarlHembrough: oops - fixed now 👍

PR #94, opened 2018-05-25
Feature/logging request ids

nshumoogum:
### What

Additional functions in the following package handlers, log and audit to enable the capture of request ids, user and caller identities from go context.

### How to review

Check unit tests pass foreach of those packages described above and logical changes make sense.

It might be worth reviewing in order of the three commits to understand the changes.

### Who can review

Team B


handlers/requestID/handler.go
```diff
@@ -11,22 +11,18 @@ var requestIDRandom = rand.New(rand.NewSource(time.Now().UnixNano()))
 
 type contextKey string
 
+// ContextKey represents the constant key name name
```
mattrout92: key name name?

handlers/requestID/handler.go
```diff
@@ -36,7 +32,24 @@ func Handler(size int) func(http.Handler) http.Handler {
...
+
+// Get retrieves the value of the context key (request-id)
 func Get(ctx context.Context) string {
 	id, _ := ctx.Value(ContextKey).(string)
 	return id
 }
+
+// Set creates/overwrites the value of the context key (request-id)
+func Set(ctx context.Context, requestID string) context.Context {
+	ctx = context.WithValue(ctx, ContextKey, requestID)
```
mattrout92: just return:

```go
return context.WithValue(ctx, ContextKey, requestID)
```

log/log.go
```diff
@@ -179,6 +182,12 @@ func ErrorC(context string, err error, data Data) {
...
 
+// ErrorCtx is a structured info message for go context
+func ErrorCtx(ctx context.Context, err error, data Data) {
+	context := requestID.Get(ctx)
+	ErrorC(context, err, data)
+}
+
 // ErrorR is a structured error message for a request
 func ErrorR(req *http.Request, err error, data Data) {
 	ErrorC(Context(req), err, data)
```
mattrout92: consider renaming the "Context" function since it has very little to do with a go context.Context

gedge: yeah, how about `correlationId`

log/log.go
```diff
@@ -33,8 +36,8 @@ func configureHumanReadable() {
 // Data contains structured log data
 type Data map[string]interface{}
 
-// Context returns a context ID from a request (using X-Request-Id)
-func Context(req *http.Request) string {
+// GetRequestID returns the request ID from a request (using X-Request-Id)
+func GetRequestID(req *http.Request) string {
 	return req.Header.Get("X-Request-Id")
```
mattrout92: It would probably better if "X-Request-Id" was a constant in the `common` package, since it is used across multiple packages

audit/log_test.go
```diff
@@ -0,0 +1,205 @@
...
+		eventData = log.Data{}
+	})
+}
+
+func TestAuditInfoLog(t *testing.T) {
+
+	var eventName, eventCorrelationKey string
+	var eventData log.Data
+	log.Event = func(name string, correlationKey string, data log.Data) {
+		eventName = name
```
gedge: the tests look long & opaque to me,  I'd like to see use of a struct:
e.g.
```
// top of package
type eventResults struct { name string, cKey string, data log.Data }
...
// above this func
var eventResult eventResults
...
// in this func
eventResult = eventResults{name: name, cKey: correlationKey, data: data}
...
// in tests...
So(eventResult, ShouldResemble, eventResults{
          name: "info", 
          cKey: "",
          data: log.Data{message:"info message"},
})
...
// at the end of each test (also fixes the problem that you're not resetting e.g. eventName)
eventResult = eventResults{}
```
?

log/log.go
```diff
@@ -115,24 +119,24 @@ func event(name string, context string, data Data) {
...
 		b, _ = json.Marshal(map[string]interface{}{
-			"created":   time.Now(),
-			"event":     "log_error",
-			"namespace": Namespace,
-			"context":   context,
-			"data":      map[string]interface{}{"error": err.Error()},
+			"created":        time.Now(),
+			"event":          "log_error",
+			"namespace":      Namespace,
+			"correlationKey": correlationKey,
```
gedge: snake case, please

handlers/requestID/handler.go
```diff
@@ -36,7 +32,24 @@ func Handler(size int) func(http.Handler) http.Handler {
 	}
 }
 
+// NewRequestID generates a random string of requested length
+func NewRequestID(size int) string {
+	var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
```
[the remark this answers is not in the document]
nshumoogum: because type is []rune this would have to be a package level var not const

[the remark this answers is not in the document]
gedge: yeah, that did cross my mind - var is fine

common/identity.go
```diff
@@ -5,14 +5,17 @@ import (
...
+// A list of common constants used across go-ns packages
 const (
 	FlorenceHeaderKey        = "X-Florence-Token"
 	DownloadServiceHeaderKey = "X-Download-Service-Token"
 
-	AuthHeaderKey = "Authorization"
-	UserHeaderKey = "User-Identity"
+	AuthHeaderKey    = "Authorization"
+	UserHeaderKey    = "User-Identity"
+	RequestHeaderKey = "X-Request-Id"
```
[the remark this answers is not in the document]
nshumoogum: and the handler tests too :/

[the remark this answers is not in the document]
gedge: I tend not to use consts like that in tests, so that the tests ensure that the consts don't end up being "" or some other brokenness leaks into the test

audit/log_test.go
```diff
@@ -0,0 +1,194 @@
...
+	Convey("with empty context", t, func() {
+		LogError(ctx, errTest, nil)
+
+		So(eventRes, ShouldResemble, eventResults{
+			name: "error",
+			cKey: "",
+			data: log.Data{"message": "test error", "error": errTest},
+		})
+
+		eventRes = eventResults{}
```
gedge: I don't think this is necessary, now.

audit/log_test.go
```diff
@@ -0,0 +1,194 @@
...
+func TestInfoLog(t *testing.T) {
+
+	var eventRes eventResults
+	log.Event = func(name string, correlationKey string, data log.Data) {
+		eventRes = eventResults{
+			name: name,
+			cKey: correlationKey,
+			data: data,
+		}
+	}
```
gedge: this is all repeated from the previous test, could they be moved to package-level?

PR #95, opened 2018-05-29
move correlation-id into common pkg; auto append same to rchttp requests

gedge:
### What

- make the `context`/`requestId`/correlation-id utility functions more available by moving them to the `common` package
- get `rchttp` to add the `x-request-id` header to outbound requests (using, and adding to, any existing correlation-id in the `context`)

### How to review

You can run
- a listener in one window: `nc -l localhost 8081`
- the example `go run cmd/rchttp-example/main.go` in another window (see use of `common.WithRequestId`)

see the header appear in the `nc` window

### Who can review

Team B


rchttp/client_test.go
```diff
@@ -148,15 +148,46 @@ func TestClientNoRetries(t *testing.T) {
...
+
+			call, err := unmarshallResp(resp)
+			So(err, ShouldBeNil)
+
+			Convey("Then the server sees the auth and user headers", func() {
+				So(call.CallCount, ShouldEqual, expectedCallCount)
+				So(call.Method, ShouldEqual, "GET")
+				So(call.Error, ShouldEqual, "")
+				So(call.Headers[common.AuthHeaderKey], ShouldResemble, []string{common.BearerPrefix + expectedAuthToken})
+				So(call.Headers[common.UserHeaderKey], ShouldResemble, []string{testUser})
```
nshumoogum: Should there be an assertion to check requestID (the upstream request ID) ?

gedge: done

PR #96, opened 2018-05-29
Feature/mongo timestamp helpers

gedge:
### What

add utility functions for handling mongo timestamps

### How to review

check the tests, deploy the functions

### Who can review

team b


mongo/healthcheck.go
```diff
@@ -2,7 +2,7 @@ package mongo
 
 import (
 	"github.com/ONSdigital/go-ns/log"
-	"github.com/globalsign/mgo"
+	mgo "github.com/gedge/mgo"
```
mattrout92: When I attempt to `go get github.com/gedge/mgo` I get the following error:

gedge/mgo/auth.go:38:2: use of internal package not allowed

mongo/mongo.go
```diff
@@ -53,3 +66,73 @@ func Close(ctx context.Context, session *mgo.Session) error {
...
+	var currentDate bson.M
+	var ok bool
+	if currentDate, ok = updateDoc["$currentDate"].(bson.M); !ok {
+		currentDate = bson.M{}
+	}
+	switch v := val.(type) {
+	case bool, bson.M:
+		currentDate[key] = v
+	default:
+		panic("Cannot handle that type")
```
mattrout92: In don't think this should panic

mongo/mongo_test.go
```diff
@@ -23,18 +23,42 @@ var (
...
+type TestModel struct {
+	State           string              `bson:"state"`
+	NewKey          int                 `bson:"new_key,omitempty"`
+	LastUpdated     time.Time           `bson:"last_updated"`
+	UniqueTimestamp bson.MongoTimestamp `bson:"unique_timestamp,omitempty"`
+}
+
+type Times struct {
+	LastUpdated     time.Time           `bson:"last_updated"`
+	UniqueTimestamp bson.MongoTimestamp `bson:"unique_timestamp,omitempty"`
```
nshumoogum: This functionality is not needed, we only need a unique timestamp at the document level, not in nested objects.

mongo/mongo.go
```diff
@@ -5,9 +5,22 @@ import (
...
 
+// keep these in sync with Timestamps tags below
+const (
+	lastUpdatedKey     = "last_updated"
+	uniqueTimestampKey = "unique_timestamp"
+)
+
+// keep tags in sync with above const
+type Timestamps struct {
+	LastUpdated     time.Time           `bson:"last_updated,omitempty"     json:"last_updated,omitempty"`
```
mattrout92: If you wish to omitempty on this then LastUpdated must become a pointer, as there is no `empty` value for a time.Time (only a zero value)

mongo/healthcheck.go
gedge: yes, I think that's because I put my changes on a different branch: `feature/mongotimestamp-funcs-n-imports`
to avoid changing all the `import` statements on `master`
not sure how to tell `go get` to get a particular branch

mongo/mongo.go
```diff
@@ -5,9 +5,22 @@ import (
 	"errors"
 	"time"
 
-	"github.com/globalsign/mgo"
+	mgo "github.com/gedge/mgo"
+	"github.com/gedge/mgo/bson"
```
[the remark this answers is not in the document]
gedge: I don't agree, I think it's best to keep the whole repo together

mongo/mongo_test.go
gedge: that struct is only for the test

mongo/mongo.go
gedge: agreed, updating...

mongo/mongo.go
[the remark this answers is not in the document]
gedge: agreed, updating...

gedge: amending...

mongo/mongo.go
```diff
@@ -17,8 +17,8 @@ const (
 
 // keep tags in sync with above const
 type Timestamps struct {
-	LastUpdated     time.Time           `bson:"last_updated,omitempty"     json:"last_updated,omitempty"`
-	UniqueTimestamp bson.MongoTimestamp `bson:"unique_timestamp,omitempty" json:"-"`
+	LastUpdated     time.Time            `bson:"last_updated,omitempty"     json:"last_updated,omitempty"`
```
nshumoogum: Should this also be a pointer or remove omitempty from bson and json tags?

gedge: `time.Time` has a _zero value_ which is considered sufficient (and recommends using `mydoc.LastUpdated.IsZero()`)

from https://golang.org/pkg/time/#Time

PR #97, opened 2018-05-30
add new clients/importapi for Import API client stuff

gedge:
### What

started a client for the import api (currently only `GET /jobs/jobId` and `PUT /jobs/jobId` for `state` changes)

### How to review

run the test

### Who can review

team b

### DIFF
--- a/clients/importapi/importapi.go
+++ b/clients/importapi/importapi.go
@@ -81,7 +81,7 @@ func (c *Client) Healthcheck() (string, error) {
 // ImportJob comes from the Import API and links an import job to its (other) instances
 type ImportJob struct {
 	JobID string  `json:"id"`
-	Links LinkMap `json:"links,ignoreempty"`
+	Links LinkMap `json:"links,omitempty"`
 }
 
 // LinkMap is an array of instance links associated with am import job
@@ -100,7 +100,7 @@ func (api *Client) GetImportJob(ctx context.Context, importJobID string) (Import
 	var importJob ImportJob
 	path := api.url + "/jobs/" + importJobID
 
-	jsonBody, httpCode, err := api.getJson(ctx, path, 0, nil)
+	jsonBody, httpCode, err := api.getJSON(ctx, path, 0, nil)
 	if httpCode == http.StatusNotFound {
 		return importJob, false, nil
 	}
@@ -137,7 +137,7 @@ func (api *Client) UpdateImportJobState(ctx context.Context, jobID string, newSt
 	path := api.url + "/jobs/" + jobID
 	jsonUpload := []byte(`{"state":"` + newState + `"}`)
 
-	jsonResult, httpCode, err := api.putJson(ctx, path, 0, jsonUpload)
+	jsonResult, httpCode, err := api.putJSON(ctx, path, 0, jsonUpload)
 	logData := log.Data{
 		"path":        path,
 		"importJobID": jobID,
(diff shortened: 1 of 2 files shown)


clients/importapi/importapi.go
```diff
@@ -0,0 +1,246 @@
...
+		err = errors.New("Bad HTTP response")
+	}
+	if err != nil {
+		log.ErrorC("UpdateImportJobState", err, logData)
+		return err
+	}
+	return nil
+}
+
+func (api *Client) getJson(ctx context.Context, path string, attempts int, vars url.Values) ([]byte, int, error) {
```
eldeal: Consistent casing for acronyms preferred https://github.com/golang/go/wiki/CodeReviewComments#initialisms

clients/importapi/importapi.go
```diff
@@ -0,0 +1,246 @@
...
+		"httpCode":    httpCode,
+		"jsonBody":    string(jsonBody),
+	}
+	var isFatal bool
+	if err == nil && httpCode != http.StatusOK {
+		if httpCode < http.StatusInternalServerError {
+			isFatal = true
+		}
+		err = errors.New("Bad response while getting import job")
+	} else {
```
eldeal: Dont think the else is required, not even sure the isFatal is actually adding anything. Where the err != nil, it will be caught by the following statement, and where the status is actually OK we probably don't want to set it to true? In either case its not being used as in the last return, so why not just directly return true in the `if err != nil`

Possbily 122 should be `if err != nil  || isFatal` to capture the correct cases?

clients/importapi/importapi.go
```diff
@@ -0,0 +1,246 @@
...
+
+func (api *Client) getJson(ctx context.Context, path string, attempts int, vars url.Values) ([]byte, int, error) {
+	return callJsonAPI(ctx, api.client, "GET", path, vars)
+}
+
+func (api *Client) putJson(ctx context.Context, path string, attempts int, payload []byte) ([]byte, int, error) {
+	return callJsonAPI(ctx, api.client, "PUT", path, payload)
+}
+
+// func callJsonAPI(ctx context.Context, client *rchttp.Client, method, path string, payload interface{}) ([]byte, int, error) {
```
eldeal: do we need both/either commented strings here? If they're for future maybe add a TODO

clients/importapi/importapi.go
```diff
@@ -0,0 +1,246 @@
...
+// NewAPIResponse creates an error response, optionally adding body to e when status is 404
+func NewAPIResponse(resp *http.Response, uri string) (e *ErrInvalidAPIResponse) {
+	e = &ErrInvalidAPIResponse{
+		actualCode: resp.StatusCode,
+		uri:        uri,
+	}
+	if resp.StatusCode == http.StatusNotFound {
+		body, err := getBody(resp)
+		if err != nil {
+			e.body = "Client failed to read Client body"
```
eldeal: failed to read `client` or `response` body?

clients/importapi/importapi.go
```diff
@@ -0,0 +1,246 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"github.com/ONSdigital/go-ns/common"
+	"github.com/ONSdigital/go-ns/log"
+	"github.com/ONSdigital/go-ns/rchttp"
+)
+
+const service = "import-api"
```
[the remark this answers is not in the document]
gedge: `cd go-ns && ack '\bservice\s+=\s+' clients` would disagree (these are clients, so perhaps should be using the remote service name, not the repo name?)

clients/importapi/importapi.go
```diff
@@ -0,0 +1,246 @@
...
+		return service, NewAPIResponse(resp, "/healthcheck")
+	}
+
+	return service, nil
+}
+
+// ImportJob comes from the Import API and links an import job to its (other) instances
+type ImportJob struct {
+	JobID string  `json:"id"`
+	Links LinkMap `json:"links,ignoreempty"`
```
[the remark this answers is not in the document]
gedge: looks like a fubar!  good spot

clients/importapi/importapi.go
gedge: 👍

clients/importapi/importapi.go
gedge: both routes of this `if` will ensure `err` is set and hence trigger the follow (`if err != nil {`)
*but* only one of those routes will leave `isFatal == false` (when `httpCode >= 500` server error or worse - i.e. may be worth retrying)

any other/real error sets `isFatal` to `true`

hope that clarifies

(though, it does look unclear, I admit)

clients/importapi/importapi.go
gedge: oops, that's my bad, another good spot

clients/importapi/importapi.go
gedge: 👍

PR #104, opened 2018-06-07
Allow florence token to be set on api clients

mattrout92:
### What

- Allow florence token to be added to context (for frontend controllers to pass to apis)
- Update clientlogs to take in context

### How to review

Verify logic correct

### Who can review

Anyone

### DIFF
--- a/common/identity.go
+++ b/common/identity.go
@@ -25,9 +25,10 @@ const (
 	LegacyUser           = "legacyUser"
 	BearerPrefix         = "Bearer "
 
-	UserIdentityKey   = ContextKey("User-Identity")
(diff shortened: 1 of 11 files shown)


common/identity.go
```diff
@@ -47,6 +48,12 @@ func IsUserPresent(ctx context.Context) bool {
 
 }
 
+// IsFlorenceIdentityPresent determines if a florence identity is present on the given context
+func IsFlorenceIdentityPresent(ctx context.Context) bool {
```
daiLlew: Just wonder if it might be useful to also return the token if it exists?

mattrout92: I should probably have created a separate function to retrieve the token from context

PR #108, opened 2018-06-18
Add helper function to drain HTTP request bodies

CarlHembrough:
### What

Add helper function to drain HTTP request bodies

### How to review

Review code - see / test usage in Filter API PR: https://github.com/ONSdigital/dp-filter-api/pull/101

### Who can review

Anyone

### DIFF
--- a/request/util.go
+++ b/request/util.go
@@ -1,0 +1,24 @@
+package request
+
+import (
+	"github.com/ONSdigital/go-ns/log"
+	"io"
+	"io/ioutil"
+	"net/http"
+)
+
+// DrainBody drains the body of the given of the given HTTP request.
+// any handler function that reads the request body should ensure that this
+// function is called.
+// ``
+func DrainBody(r *http.Request) {
+	_, err := io.Copy(ioutil.Discard, r.Body)
+	if err != nil {
+		log.InfoCtx(r.Context(), "error draining request body", nil)
+	}
+
+	err = r.Body.Close()
+	if err != nil {
+		log.InfoCtx(r.Context(), "error closing request body", nil)
+	}
+}


request/util.go
```diff
@@ -0,0 +1,24 @@
...
+func DrainBody(r *http.Request) {
+	_, err := io.Copy(ioutil.Discard, r.Body)
+	if err != nil {
+		log.InfoCtx(r.Context(), "error draining request body", nil)
+	}
+
+	err = r.Body.Close()
+	if err != nil {
+		log.InfoCtx(r.Context(), "error closing request body", nil)
+	}
```
gedge: why not return errors?

request/util.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/ONSdigital/go-ns/log"
+	"io"
+	"io/ioutil"
+	"net/http"
+)
+
+// DrainBody drains the body of the given of the given HTTP request.
+// any handler function that reads the request body should ensure that this
```
gedge: "any" seems to assume that none will (or should) drain it themselves - strong words

request/util.go
gedge: hmm, I see the caller is in a `defer` - which answers that question

CarlHembrough: I was unsure whether to return errors from this function. I realise that typically errors are not returned from lib's, though this seemed to sit somewhere between. I wrote the function and then decided to put it in go-ns for reuse. I could return errors and have them handled in a deferred function, though it would make the code more verbose. What do you think?

request/util.go
CarlHembrough: I have removed this line - I will leave it to the caller to decide if it's required or not 😄

request/util.go
gedge: It's likely to be in a `defer` and we are discarding the body, so I think your first instinct - to log and ignore - is best

PR #114, opened 2018-06-27
healthcheck: recoveryDuration, stateChangeChan, requestCheckChan

gedge:
### What

As discussed, an update to healthchecking for quicker recovery, channels for state changes, check requests

### How to review

check the code and tests

### Who can review

team b

### DIFF
--- a/healthcheck/ticker.go
+++ b/healthcheck/ticker.go
@@ -12,6 +12,7 @@ type Ticker struct {
 	timeTicker *time.Ticker
 	asyncCheck chan bool
 	closing    chan bool
+	closed     chan bool
 }
 
 // NewTicker returns a new ticker that checks the given clients at intervals of the given duration
@@ -21,8 +22,10 @@ func NewTicker(duration, recoveryDuration time.Duration, clients ...Client) *Tic
 
 // NewTickerWithAlerts returns a new ticker that checks the given clients at intervals
 // - intervals vary in length (when healthy, use `duration`; at startup, or unhealthy state, use `recoveryDuration`)
-// - sends alerts of any change in health via the given channel (if non-nil) - with the new state (i.e. `true` if newly-healthy)
-// - healthchecks can be requested using requestCheckChan (sending `false` on this channel indicates that a failure has prompted this check)
+// - sends alerts of any change in health via the given channel (if non-nil) - with the new state (`true`:newly-healthy, `false`:newly-unhealthy)
+// - healthchecks can be requested using `requestCheckChan`
+//   - sending `false` on this channel indicates that a failure has prompted this check (fail fast)
+//   - sending `true` indicates that a success has happened (unlikely to indicate 100% health)
 // - healthchecks will only be run at a maximum of `recoveryDuration` frequency
 func NewTickerWithAlerts(
 	duration,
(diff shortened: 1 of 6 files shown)


healthcheck/ticker.go
```diff
@@ -1,35 +1,135 @@
...
+// NewTicker returns a new ticker that checks the given clients at intervals of the given duration
+func NewTicker(duration, recoveryDuration time.Duration, clients ...Client) *Ticker {
+	return NewTickerWithAlerts(duration, recoveryDuration, nil, nil, clients...)
 }
 
-// NewTicker returns a new ticker that checks the given clients at intervals of the given duration.
-func NewTicker(duration time.Duration, clients ...Client) *Ticker {
+// NewTickerWithAlerts returns a new ticker that checks the given clients at intervals
+// - intervals vary in length (when healthy, use `duration`; at startup, or unhealthy state, use `recoveryDuration`)
+// - sends alerts of any change in health via the given channel (if non-nil) - with the new state (i.e. `true` if newly-healthy)
```
CarlHembrough: I assume it will be false if it's newly not healthy? might be worth explicitly stating this

healthcheck/ticker.go
```diff
@@ -1,35 +1,135 @@
...
+// - healthchecks can be requested using requestCheckChan (sending `false` on this channel indicates that a failure has prompted this check)
```
CarlHembrough: what else would prompt this check? is there a difference between the check thats triggered via false and true?

healthcheck/ticker.go
```diff
@@ -1,35 +1,135 @@
...
 
-		for range timeTicker.C {
+			} else {
 
-			log.Debug("conducting service healthcheck", nil)
-			MonitorExternal(clients...)
+				lastCheckStarted = now
+
+				// run check in the background,  mutexChecking.Lock() applies
+				go func() {
```
CarlHembrough: I cannot see why this go routine is required. It seems we are only allowing one healthcheck to run at a time, so can we do the check in the parent go routine?

healthcheck/ticker.go
```diff
@@ -1,35 +1,135 @@
...
-	return &Ticker{
-		timeTicker: timeTicker,
-	}
+	return &ticker
 }
 
 // Close the ticker to exit its internal goroutine
 func (ticker *Ticker) Close() {
 	ticker.timeTicker.Stop()
+	close(ticker.closing)
```
CarlHembrough: Should we block until it has actually closed the ticker go routine?

healthcheck/ticker.go
gedge: you are right, I will add clarification

healthcheck/ticker.go
gedge: effectively, no difference (just that of the expected _current health state_).
but I'll expand the comment (I imagine fewer reasons for `true` to be used, perhaps in my scenario - kafka-backoffs - but may happen in, say, a web server, where non-green retries are more likely)

healthcheck/ticker.go
gedge: If `MonitorExternal` is running, and taking a minute to timeout for each external service (so can be taking several minutes), then the goroutine allows the app to send `requestCheckChan` messages to downgrade the health (fail fast) _despite this blocking_ - kind of belts-and-braces.

I suppose you could argue that there should be a separate goroutine for that listener, but this makes sense to me. Does that help? 🤔

CarlHembrough: I see - that makes sense 👍

PR #117, opened 2018-07-09
Feature/copy audit parameters

nshumoogum:
### What

Create copy method for audit parameters, this will allow mocked auditors to not have their parameters overwritten every time a new audit message is created; which can then allow one to test what parameters get set when.

Also renamed package name `audit_mock` to `testaudit`

### How to review

Check code changes make sense

### Who can review

Describe who worked on the changes, so that other people can review.


audit/auditortest/README.md
```diff
@@ -1,6 +1,6 @@
 ### Testing auditing
 
-The `audit_mocks` package provides useful methods for verifying calls to `audit.Record()` reducing the amount of 
+The `auditortest` package provides useful methods for verifying calls to `audit.Record()` reducing the amount of
```
gedge: this change needs to be reflected in several places below

PR #119, opened 2018-07-10
Add audit process to identity check

nshumoogum:
### What

Add audit process to the identity check wrapper and improve logging in the authentication middleware.

### How to review

Check changes make sense and all unit tests pass.

### Who can review

Anyone

### DIFF
--- a/audit/helper.go
+++ b/audit/helper.go
@@ -1,0 +1,51 @@
+package audit
+
+import (
+	"context"
+	"strings"
+
+	"github.com/ONSdigital/go-ns/common"
+)
+
+// pathIDs relate the possible string matches in path with the respective
+// parameter name as key values pairs
+var pathIDs = map[string]string{
+	"jobs":                  "job_id",
+	"datasets":              "dataset_id",
+	"editions":              "edition",
+	"versions":              "version",
+	"dimensions":            "dimension",
+	"options":               "option",
+	"instances":             "instance_id",
+	"inserted_observations": "inserted_observations",
+	"node_id":               "node_id",
+	"filters":               "filter_blueprint_id",
+	"filter-outputs":        "filter_output_id",
+	"hierarchies":           "instance_id",
+	// TODO add params for codelist API endpoints
+}
+
+// GetParameters populates audit parameters with path variable values
+func GetParameters(ctx context.Context, path string) common.Params {
+	auditParams := common.Params{}
+
+	callerIdentity := common.Caller(ctx)
+	if callerIdentity != "" {
+		auditParams["caller_identity"] = callerIdentity
+	}
+
+	splitPath := strings.Split(path, "/")
+
+	for i, pathValue := range splitPath {
+		if i+1 >= len(splitPath) {
+			break
+		}
+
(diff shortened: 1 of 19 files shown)


clients/identity/authentication.go
```diff
@@ -55,7 +60,7 @@ func (api IdentityClient) CheckRequest(req *http.Request) (context.Context, int,
 
 	// if neither user nor service request, return unchanged ctx
 	if !isUserReq && !isServiceReq {
-		return ctx, http.StatusUnauthorized, nil
+		return ctx, http.StatusUnauthorized, errors.WithMessage(errUnableToIdentifyRequest, "no headers set on request")
```
gedge: I'm a little concerned that "not providing auth headers" is considered an error.
I'm happy with `StatusUnauthorized` being returned, but why an error _too_ - shouldn't errors be reserved for failures, not for negative checks?

audit/helper.go
```diff
@@ -0,0 +1,51 @@
...
+
+	for i, pathValue := range splitPath {
+		if i+1 >= len(splitPath) {
+			break
+		}
+
+		parameter, ok := pathIDs[pathValue]
+		if ok {
+			auditParams[parameter] = splitPath[i+1]
+		}
```
gedge: how about adding else { auditParams[pathValue] = splitPath[i+1] } (i.e. default to pathValue if that key doesn't exist in the map - and then you can remove the 2 key-value pairs above where they match (and future expansion defaults to using the existing pathValue and doesn't go undetected)?

nshumoogum: We can't assume the structure of the url will always be the same;

Most of our urls are like so:

`/<lists>/<id>/<lists>/<id>.....`

But there are some different structures like the following:

`/search/datasets/<id>` and `/hierarchies/<instance>/<dimension>`

It is difficult to capture all these without creating a list; a map is used to know when an id is a `dataset_id`, `instance_id` etc.

clients/identity/authentication.go
```diff
@@ -26,19 +26,22 @@ type Client common.APIClient
 
 // Clienter provides an interface to checking identity of incoming request
 type Clienter interface {
-	CheckRequest(req *http.Request) (context.Context, int, error)
+	CheckRequest(req *http.Request) (context.Context, int, authError, httpError)
```
gedge: hate to be pedantic (...), but can you reconsider these errors for clarity, please: how about
`..., authFailure, error)`
i.e.:
- keep `error` for *general* http/json errors (serious, general, external errors)
- use `authFailure` (type `error`) to indicate whether auth failed

the crucial problem, for me, being that your `httpError` can also results from a json marshall error (which _is_ http related, but seems more general to me), but using just `error` emphasises a systemic failure rather than a semantic (authFailure) issue.  Idiomatic?

clients/identity/authentication.go
```diff
@@ -26,19 +26,22 @@ type Client common.APIClient
...
 }
 
-// NewAPIClient returns an IdentityClient
+// NewAPIClient returns an Client
```
gedge: `an` => `a`

PR #121, opened 2018-07-12
Update dataset API client for missing metadata fields and new download type

eldeal:
### What

Update dataset API client to make it work for CSVW generation

 I wouldn't be surprised if these changes break things in other places, so by all means highlight those and I'll make appropriate changes. Otherwise going to rely on vendoring to prevent this causing issues.

### How to review

Check changes make sense

Related PRs:
https://github.com/ONSdigital/dp-dataset-api/pull/162
https://github.com/ONSdigital/dp-dataset-exporter/pull/46
https://github.com/ONSdigital/dp-download-service/pull/33

### Who can review

Anyone


clients/dataset/data.go
```diff
@@ -85,6 +93,12 @@ type Publisher struct {
 	Type string `json:"type"`
 }
 
+// UsageNote represents a note containing extra information associated to the resource
+type UsageNote struct {
+	Title string `json:"title,omitempty"`
+	Note  string `json:"note,omitempty"`
```
nshumoogum: Me being pedantic but can we have this in alphabetical order :p

PR #125, opened 2018-07-16
Drain request body in identity middleware

CarlHembrough:
### What

For cases where some failure occurs in the identity middleware the
request body is not drained. This PR adds a call to drain the request
body when an error occurs and the downstream handler is not called.
 
I have updated the requests in existing tests to contain a request body (regardless of HTTP method) allowing us to assert that the body is drained.

### How to review

Review code / test

### Who can review

Anyone

### DIFF
--- a/identity/identity_test.go
+++ b/identity/identity_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ONSdigital/go-ns/common/commontest"
 	"github.com/pkg/errors"
 	. "github.com/smartystreets/goconvey/convey"
+	"io"
 )
 
 const (
@@ -30,7 +31,7 @@ func TestHandler_NoHeaders(t *testing.T) {
 
 	Convey("Given a http request with no headers", t, func() {
 
-		req := httptest.NewRequest("GET", url, nil)
+		req := httptest.NewRequest("GET", url, bytes.NewBufferString("some body content"))
 		responseRecorder := httptest.NewRecorder()
 
 		httpClient := &commontest.RCHTTPClienterMock{}
@@ -54,6 +55,11 @@ func TestHandler_NoHeaders(t *testing.T) {
 			Convey("Then the http response should have a 401 status", func() {
 				So(responseRecorder.Result().StatusCode, ShouldEqual, http.StatusUnauthorized)
 			})
+
+			Convey("Then the request body has been drained", func() {
+				_, err := req.Body.Read(make([]byte, 1))
+				So(err, ShouldEqual, io.EOF)
+			})
 		})
 	})
 }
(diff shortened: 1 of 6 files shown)


identity/identity_test.go
```diff
@@ -154,7 +170,7 @@ func TestHandler_florenceToken(t *testing.T) {
 
 	Convey("Given a request with a florence token, and mock client that returns 200", t, func() {
 
-		req := httptest.NewRequest("GET", url, nil)
+		req := httptest.NewRequest("GET", url, bytes.NewBufferString("some body content"))
```
gedge: forgot to check for draining within this test?

identity/identity_test.go
```diff
@@ -266,7 +287,7 @@ func TestHandler_authToken(t *testing.T) {
 
 	Convey("Given a request with an auth token, and mock client that returns 200", t, func() {
 
-		req := httptest.NewRequest("GET", url, nil)
+		req := httptest.NewRequest("GET", url, bytes.NewBufferString("some body content"))
```
gedge: missing drain check?

identity/identity_test.go
```diff
@@ -328,7 +349,7 @@ func TestHandler_bothTokens(t *testing.T) {
 
 	Convey("Given a request with both a florence token and service token", t, func() {
 
-		req := httptest.NewRequest("GET", url, nil)
+		req := httptest.NewRequest("GET", url, bytes.NewBufferString("some body content"))
```
gedge: missing drain check?

request/request_test.go
```diff
@@ -0,0 +1,44 @@
...
+				_, err = r.Body.Read(make([]byte, 1))
+				So(err, ShouldEqual, io.EOF)
+			})
+		})
+	})
+}
+
+func TestDrainBody_WithoutRequestBody(t *testing.T) {
+
+	Convey("Given a request without a nil body", t, func() {
```
gedge: `with a nil body`?  worth adding a test for *no body*, too, methinks

identity/identity_test.go
CarlHembrough: I was debating if I should add a test in this scenario. The downstream handler is called, so the request body should not be drained in this case. I have added assert to check that it was not drained.

identity/identity_test.go
CarlHembrough: same as above

identity/identity_test.go
CarlHembrough: same as above

request/request_test.go
CarlHembrough: Good point - I have added a test for this case

PR #126, opened 2018-07-24
add vault kv-v2 (versioned) functionality, with example

gedge:
### What

add helper methods for vault kv engine v2 (versioned keys)

### How to review

see the example code, check it, run it

### Who can review

team awesome! 🤘

### DIFF
--- a/vault/client.go
+++ b/vault/client.go
@@ -1,10 +1,13 @@
 package vault
 
 import (
+	"encoding/json"
 	"errors"
 
 	vaultapi "github.com/hashicorp/vault/api"
 )
+
+var ErrorKeyNotFound = errors.New("key not found")
 
 // VaultClient Used to read and write secrets from vault
 type VaultClient struct {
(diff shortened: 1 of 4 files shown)


vault/client.go
```diff
@@ -1,11 +1,14 @@
 package vault
 
 import (
+	"encoding/json"
 	"errors"
 
 	vaultapi "github.com/hashicorp/vault/api"
 )
 
+var ErrorKeyNotFound = errors.New("key not found")
```
mattrout92: Go convention prefers `ErrKeyNotFound` over`ErrorKeyNotFound`

vault/client.go
```diff
@@ -88,3 +91,47 @@ func (c *VaultClient) WriteKey(path, key, value string) error {
...
+	if err != nil {
+		return nil, -1, err
+	}
+	if len(secret) == 0 {
+		// if there is no secret and no error return a empty map
+		return secret, -1, nil
+	}
+	verObj, ok := secret["metadata"].(map[string]interface{})["version"]
+	if !ok {
+		return nil, -1, errors.New("key version not found")
```
mattrout92: For consistency it may be worth defining this error as a variable, or removing the `ErrorKeyNotFound`

vault/client.go
```diff
@@ -88,3 +91,47 @@ func (c *VaultClient) WriteKey(path, key, value string) error {
...
+func (c *VaultClient) VRead(path string) (map[string]interface{}, int64, error) {
+	secret, err := c.Read(path)
+	if err != nil {
+		return nil, -1, err
+	}
+	if len(secret) == 0 {
+		// if there is no secret and no error return a empty map
+		return secret, -1, nil
+	}
+	verObj, ok := secret["metadata"].(map[string]interface{})["version"]
```
mattrout92: Could be worth doing this cast check in two steps to avoid interface panics if running version 1

vault/client.go
gedge: done

vault/client.go
gedge: done

vault/client.go
gedge: and done

PR #140, opened 2019-05-14
added collection-id handler

mikeAdamss:
### What

1st stab at collection id handler.

Spoke with James, we're ignoring the Travis CI issue (will add a card for it)

### How to review

Sanity check.

### Who can review

Anyone.

### DIFF
--- a/handlers/collectionID/handler.go
+++ b/handlers/collectionID/handler.go
@@ -11,6 +11,11 @@ func Handler(h http.Handler) http.Handler {
 
 		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 			collectionID := req.Header.Get(common.CollectionIDHeaderKey)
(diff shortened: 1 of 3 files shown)


handlers/collectionID/handler.go
```diff
@@ -0,0 +1,16 @@
...
+	"net/http"
+	"context"
+	"github.com/ONSdigital/go-ns/common"
+)
+
+// Handler is a wrapper which adds a CollectionID header if one does not yet exist
+func Handler(h http.Handler) http.Handler {
+
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			collectionID := req.Header.Get(common.CollectionIDHeaderKey)
```
daiLlew: Only comment I have is to add an `if present` check. i.e. only add the item to the context if the header exists.

PR #142, opened 2019-05-20
added handler for collection ids from cookies

mikeAdamss:
### What

Updated collectionID handler to be two handlers, one for extracting from headers, the other from cookies.

### How to review

Sanity check.

### Who can review

Anyone.


handlers/collectionID/handler.go
```diff
@@ -1,21 +1,44 @@
...
 
-			if collectionID != "" {
-				req = req.WithContext(context.WithValue(req.Context(), common.CollectionIDHeaderKey, collectionID))
+		collectionIDCookie, err := req.Cookie(common.CollectionIDCookieKey)
+		if err == nil {
+			collectionID := collectionIDCookie.String()
+			req = req.WithContext(context.WithValue(req.Context(), common.CollectionIDHeaderKey, collectionID))
+		} else {
+			if err != http.ErrNoCookie {
+				log.ErrorCtx(req.Context(), errors.New("Unexpected error while extracting collection ID from cookie."), nil)
```
eldeal: https://github.com/ONSdigital/dp/blob/master/LOGGING_STANDARDS.md#effective-event-types logging standards suggest no period on the end of the message (i realize this is technically an error, and we may need to have a conversation about whether the standards are the same, but ive always assumed they are) and that it probably doesnt need a capital letter to start

PR #144, opened 2019-06-19
Add GetBypath method

jondewijones:
### What

Add GetBypath method

### Who can review

Anyone

### DIFF
--- a/clients/dataset/dataset.go
+++ b/clients/dataset/dataset.go
@@ -122,6 +122,54 @@ func (c *Client) Get(ctx context.Context, id string) (m Model, err error) {
 	return
 }
 
+// GetByPath returns dataset level information for a given dataset path
+func (c *Client) GetByPath(ctx context.Context, path string) (m Model, err error) {
+	uri := fmt.Sprintf("%s/%s", c.url, path)
+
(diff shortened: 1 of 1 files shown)


clients/dataset/dataset.go
```diff
@@ -122,6 +122,54 @@ func (c *Client) Get(ctx context.Context, id string) (m Model, err error) {
 	return
 }
 
+// GetByPath returns dataset level information for a given dataset path
+func (c *Client) GetByPath(ctx context.Context, path string) (m Model, err error) {
+	uri := fmt.Sprintf("%s/%s", c.url, path)
```
eldeal: might be worth sanitising the provided path and checking for preceding '/' if we're going to insert one here

clients/dataset/dataset.go
```diff
@@ -122,6 +122,54 @@ func (c *Client) Get(ctx context.Context, id string) (m Model, err error) {
...
+
+	clientlog.Do(ctx, "retrieving dataset", service, uri)
```
eldeal: we dont know that its retrieving a dataset, its just retrieving something

PR #146, opened 2019-06-26
Feature/update zebedee client to use context

jondewijones:
### What

- Add (middleware) handler for checking access token in cookie and header 
- Update Zebedee client to accept context rather than setting access token property within package

### Who can review

Anyone

### DIFF
--- a/handlers/accessToken/handler.go
+++ b/handlers/accessToken/handler.go
@@ -9,8 +9,8 @@ import (
 	"github.com/ONSdigital/go-ns/log"
 )
 
-// CheckHeader is a wrapper which adds a accessToken from the request header to context if one does not yet exist
-func CheckHeader(h http.Handler) http.Handler {
+// CheckHeaderValueAndForwardWithRequestContext is a wrapper which adds a accessToken from the request header to context if one does not yet exist
+func CheckHeaderValueAndForwardWithRequestContext(h http.Handler) http.Handler {
 
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 
@@ -24,8 +24,8 @@ func CheckHeader(h http.Handler) http.Handler {
 	})
 }
 
-// CheckCookie is a wrapper which adds a accessToken from the cookie to context if one does not yet exist
-func CheckCookie(h http.Handler) http.Handler {
+// CheckCookieValueAndForwardWithRequestContext is a wrapper which adds a accessToken from the cookie to context if one does not yet exist
+func CheckCookieValueAndForwardWithRequestContext(h http.Handler) http.Handler {
 
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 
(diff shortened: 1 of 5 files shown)


handlers/accessToken/handler.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+	"errors"
+	"net/http"
+
+	"github.com/ONSdigital/go-ns/common"
+	"github.com/ONSdigital/go-ns/log"
+)
+
+// CheckHeader is a wrapper which adds a accessToken from the request header to context if one does not yet exist
+func CheckHeader(h http.Handler) http.Handler {
```
daiLlew: Is `CheckHeader` an accurate description of what this handler actually does? `ForwardAccessTokenIfExists` or something along those lines feels more appropriate (to me anyway).

handlers/accessToken/handler.go
```diff
@@ -0,0 +1,44 @@
...
+		if accessToken != "" {
+			req = req.WithContext(context.WithValue(req.Context(), common.AccessTokenHeaderKey, accessToken))
+		}
+
+		h.ServeHTTP(w, req)
+	})
+}
+
+// CheckCookie is a wrapper which adds a accessToken from the cookie to context if one does not yet exist
+func CheckCookie(h http.Handler) http.Handler {
```
daiLlew: Again is `CheckCookie` an accurate description of what this handler does?

zebedee/client/client.go
```diff
@@ -119,19 +120,15 @@ func (c *ZebedeeClient) GetDatasetLandingPage(path string) (data.DatasetLandingP
...
+func (c *ZebedeeClient) get(ctx context.Context, path string) ([]byte, error) {
 	req, err := http.NewRequest("GET", c.zebedeeURL+path, nil)
 	if err != nil {
 		return nil, err
 	}
 
-	if len(c.accessToken) > 0 {
-		req.Header.Set(common.FlorenceHeaderKey, c.accessToken)
+	if ctx.Value("X-Florence-Token") != nil {
+		accessToken := ctx.Value("X-Florence-Token").(string)
```
daiLlew: `go-ns/common/access_token.go` defines this context key already - we should reuse it.

Also this cast could fail and if that happens it would be useful to know its failed. I am not sure if we would want to fail fast on this error but we should log that the value could not be cast as expected.

I think you can do:
```
accessToken, ok := ctx.Value("X-Florence-Token").(string)
if !ok {
  ...
}
```

zebedee/client/client.go
```diff
@@ -239,3 +236,13 @@ func (c *ZebedeeClient) GetPageTitle(uri string) (data.PageTitle, error) {
 
 	return pt, nil
 }
+
+func (c *ZebedeeClient) createRequestURL(ctx context.Context, path string) string {
+	var url string
+	if ctx.Value("Collection-Id") != nil {
```
daiLlew: Same here - use the existing context key and check the cast & log if its unsuccessful

PR #150, opened 2019-07-22
Add Welsh locale

dandwal:
### What

Added checks for language which are then added to contexts
Tests also added

### How to review

This can be checking via the router (corresponding PR: https://github.com/ONSdigital/dp-frontend-router/pull/170

### Who can review

Anyone except me

### DIFF
--- a/common/locale.go
+++ b/common/locale.go
@@ -14,6 +14,8 @@ const (
 	LocaleCookieKey = "lang"
 	LocaleHeaderKey = "LocaleCode"
 )
+var SupportedLanguages = [2]string{LangEN, LangCY}
+
 
 // SetLocaleCode sets the Locale code used to set the language
 func SetLocaleCode(req *http.Request) *http.Request {
--- a/common/locale.go
+++ b/common/locale.go
@@ -1,38 +1,43 @@
 package common
 
 import (
-	"context"
 	"net/http"
 	"strings"
 )
 
-// LocaleCode the locale code to use to determine the language to use
-const LocaleCode = ContextKey("LocaleCode")
+//LangEN is a reference to the english short-form localeCode "en"
+const LangEN = "en"
+
+//LangCY is a reference to the english short-form localeCode "cy"
+const LangCY = "cy"
 
 // SetLocaleCode sets the Locale code used to set the language
-func SetLocaleCode(req *http.Request) context.Context {
-	lang := ExtractLangFromSubDomain(req)
+func SetLocaleCode(req *http.Request) *http.Request {
+	localeCode := ExtractLangFromSubDomain(req)
 
+	// Language is overridden by cookie 'lang' here if present.
 	if c, err := req.Cookie("lang"); err == nil && len(c.Value) > 0 {
-		lang = ExtractLangFromCookie(c)
+		localeCode = ExtractLangFromCookie(c)
 	}
+	req.Header.Set("LocaleCode", localeCode)
 
-	return context.WithValue(req.Context(), LocaleCode, lang)
+	return req
 }
 
 // ExtractLangFromSubDomain returns a language based on subdomain
 func ExtractLangFromSubDomain(req *http.Request) string {
-	lang := "en"
-	if strings.Split(req.Host, ".")[0] == "cy" {
-		lang = "cy"
-	} else {
-		lang = "en"
+	args := strings.Split(req.Host, ".")
+	if len(args) == 0 {
+		// Defaulting to "en" (LangEN) if no arguments
+		return LangEN
 	}
-	return lang
+	if strings.Split(req.Host, ".")[0] == LangCY {
+		return LangCY
+	}
+	return "LangEN"
 }
 
 // ExtractLangFromCookie returns a language based on the lang cookie
 func ExtractLangFromCookie(c *http.Cookie) string {
-	lang := c.Value
-	return lang
+	return c.Value
 }
--- a/common/locale.go
+++ b/common/locale.go
@@ -1,0 +1,38 @@
+package common
+
+import (
+	"context"
+	"net/http"
+	"strings"
+)
+
+// LocaleCode the locale code to use to determine the language to use
+const LocaleCode = ContextKey("LocaleCode")
+
+// SetLocaleCode sets the Locale code used to set the language
+func SetLocaleCode(req *http.Request) context.Context {
+	lang := ExtractLangFromSubDomain(req)
+
+	if c, err := req.Cookie("lang"); err == nil && len(c.Value) > 0 {
+		lang = ExtractLangFromCookie(c)
+	}
+
+	return context.WithValue(req.Context(), LocaleCode, lang)
+}
+
+// ExtractLangFromSubDomain returns a language based on subdomain
+func ExtractLangFromSubDomain(req *http.Request) string {
+	lang := "en"
+	if strings.Split(req.Host, ".")[0] == "cy" {
+		lang = "cy"
+	} else {
+		lang = "en"
+	}
+	return lang
+}
+
+// ExtractLangFromCookie returns a language based on the lang cookie
+func ExtractLangFromCookie(c *http.Cookie) string {
+	lang := c.Value
+	return lang
+}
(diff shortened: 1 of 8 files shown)


common/locale.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// LocaleCode the locale code to use to determine the language to use
+const LocaleCode = ContextKey("LocaleCode")
+
+// SetLocaleCode sets the Locale code used to set the language
+func SetLocaleCode(req *http.Request) context.Context {
+	lang := ExtractLangFromSubDomain(req)
+
+	if c, err := req.Cookie("lang"); err == nil && len(c.Value) > 0 {
```
daiLlew: I would add a log line here to say that `lang` is coming from the cookie - useful when debugging to know exactly what is being set and from where.

common/locale.go
```diff
@@ -0,0 +1,38 @@
...
+
+// ExtractLangFromSubDomain returns a language based on subdomain
+func ExtractLangFromSubDomain(req *http.Request) string {
+	lang := "en"
+	if strings.Split(req.Host, ".")[0] == "cy" {
+		lang = "cy"
+	} else {
+		lang = "en"
+	}
+	return lang
```
daiLlew: You could simplify this:
```
if strings.Split(req.Host, ".")[0] == "cy" {
    return "cy"
}
return "en"
```
Also I know its unlikely but _potentially_ `strings.split` might return nil or empty which would cause `[0]` to nil pointer. Consider adding a defensive check:
```
args := strings.Split(req.Host, ".")
if len(args) == 0 {
   // log defaulting to en
    return "en"
}

File: common/locale.go [hunk 3]
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// ExtractLangFromCookie returns a language based on the lang cookie
+func ExtractLangFromCookie(c *http.Cookie) string {
+	lang := c.Value
```
daiLlew: Simplify to `return c.Value`

common/locale.go
```diff
@@ -0,0 +1,38 @@
+package common
+
+import (
+	"context"
+	"net/http"
+	"strings"
+)
+
+// LocaleCode the locale code to use to determine the language to use
+const LocaleCode = ContextKey("LocaleCode")
```
daiLlew: I would consider making a Lang const:

```
const LangEN = "en"
const LangCY = "cy"
...
type Lang string
```
Minor comment but they are constants and IMO it's a bit nicer dealing with a type or constant that a string.

common/locale_test.go
```diff
@@ -0,0 +1,120 @@
...
+	})
+	Convey("Given a cookie with lang set to 'cy' when ExtractLangFromCookie is called", t, func() {
+		cookie := http.Cookie{Name: "lang", Value: "en"}
+		lang := "en"
+		languageToUse := ExtractLangFromCookie(&cookie)
+
+		Convey("Then lang returns a string 'en'", func() {
+			So(languageToUse, ShouldEqual, lang)
+		})
+	})
```
daiLlew: Add another test case where the cookie value is something other than `en` or `cy` and verify it uses the `en` default.

common/locale.go
```diff
@@ -0,0 +1,38 @@
...
+
+// SetLocaleCode sets the Locale code used to set the language
+func SetLocaleCode(req *http.Request) context.Context {
+	lang := ExtractLangFromSubDomain(req)
+
+	if c, err := req.Cookie("lang"); err == nil && len(c.Value) > 0 {
+		lang = ExtractLangFromCookie(c)
+	}
+
+	return context.WithValue(req.Context(), LocaleCode, lang)
```
daiLlew: I would also consider pulling this out into it's own function - as there could be cases where we might want to use this directly.

```
// using the Lang type I mentioned in my earlier comment
func AddLangToContext(ctx context.Context, lang Lang) context.Context {
...
}

File: common/locale.go (hunk 1 above)
Comment (dandwal) [reply common/locale.go]: Added

File: common/locale.go (hunk 2 above)
Comment (dandwal) [reply common/locale.go]: Ah yes, I forget this is a go best practice thing, multiple returns in a function, I have built up a habit of avoiding multiple returns in a function as then logic flow can be directed by many returns of a function and can be hard to know when debugging that function x actually breaks out given the many situations it can be called. 

Will conform to Go best practices though :D, so thanks for pointing this out.

Also added the arguments check.

File: common/locale.go (hunk 3 above)
Comment (dandwal) [reply common/locale.go]: Refactored

File: common/locale.go (hunk 4 above)
Comment (dandwal) [reply common/locale.go]: Absolutely, shall replace the magic strings for the constants, also beneficial if we replace the localeCode to 'en_GB', 'en-GB', 'EN' or something different to "en".

File: common/locale_test.go (hunk 5 above)
Comment (dandwal) [reply common/locale_test.go]: Good shout

File: common/locale.go (hunk 6 above)
Comment (dandwal) [reply common/locale.go]: After further investigation we encountered an issue; the context was not passed in the `allRouter` handler due to being an 'http' request, not a 'rchttp' request. As such we have instead added this to a header and this line is now `req.Header.Set("LocaleCode", localeCode)`

File: common/locale.go [hunk 7]
```diff
@@ -0,0 +1,47 @@
...
+		return LangEN
+	}
+	if strings.Split(req.Host, ".")[0] == LangCY {
+		return LangCY
+	}
+	return LangEN
+}
+
+// ExtractLangFromCookie returns a language based on the lang cookie
+func ExtractLangFromCookie(c *http.Cookie) string {
```
daiLlew: 🤔I think this function name is not 100% representative of the functionality. With my über picky hat on 🤠 I think it should make clear to the caller that it will use a default if its not an accepted value. I would either:
 - Rename it to something like I think something like `GetLangCookieOrDefault` OR
 - Return an empty string if its not `en` or `cy`  and have a separate `DefaultIfBlank(lang string)` func which does the check.

Sorry if this seems like a really picky comment. But I believe that a method should do what it says and no more - its then very clear to the caller what they can expect it to do.

common/locale.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// ExtractLangFromSubDomain returns a language based on subdomain
+func ExtractLangFromSubDomain(req *http.Request) string {
+	args := strings.Split(req.Host, ".")
+	if len(args) == 0 {
+		// Defaulting to "en" (LangEN) if no arguments
+		return LangEN
+	}
+	if strings.Split(req.Host, ".")[0] == LangCY {
```
daiLlew: Use `args[0]` instead of re-doing the work.

handlers/localeCode/handler.go
```diff
@@ -0,0 +1,42 @@
...
+func CheckCookieValueAndForwardWithRequestContext(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+
+		localeCodeCookie, err := req.Cookie(common.LocaleCookieKey)
+		if err == nil {
+			localeCode := localeCodeCookie.Value
+			req = req.WithContext(context.WithValue(req.Context(), common.LocaleHeaderKey, localeCode))
+		} else {
+			if err != http.ErrNoCookie {
+				log.ErrorCtx(req.Context(), errors.New("unexpected error while extracting collection ID from cookie"), nil)
```
daiLlew: `extracting collection ID from cookie`?

handlers/localeCode/handler_test.go
```diff
@@ -0,0 +1,83 @@
...
+		Convey("when the handler is called", func() {
+			target.ServeHTTP(w, r)
+
+			Convey("then the wrapped handle is called 1 time", func() {
+				So(mockHandler.invocations, ShouldEqual, 1)
+			})
+
+			Convey("and the request context contains a value for key localeCode", func() {
+				localeCode, ok := mockHandler.ctx.Value(common.LocaleHeaderKey).(string)
+				fmt.Println("LOCALE CODE >>>>", localeCode)
```
daiLlew: remove `println`.

handlers/localeCode/handler_test.go
```diff
@@ -0,0 +1,83 @@
...
```
daiLlew: ditto

handlers/localeCode/handler.go
dandwal: Well that was an embarrassing c/p issue 😅

PR #153, opened 2019-09-02
Feature/update rchttp clients healthcheck identity

mikeAdamss:
### What

Updated healthcheck and identity clients. Created a new shared client interface for API clients but in /clients rather than /common to avoid a circular dependency (rchttp imports common).

Also needed to update the /identity middleware as it directly depends on the now altered identity client.

### How to review

Sanity check, run unit tests.

### Who can review

Anyone.


clients/baseclient.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"sync"
+	"github.com/ONSdigital/dp-rchttp"
+)
+
+// APIClient represents a common structure for any api client
+type APIClient struct {
+	BaseURL    string
+	AuthToken  string
```
eldeal: Potentially a separate issue, but shouldn't this not have AuthToken in it?

mikeAdamss: no literally the thing I was trying to get rid of, just copy pasted without thinking - ty, good spot.

PR #155, opened 2019-09-04
update dataset-ap client to new rchttp lib

mikeAdamss:
### What

Updated client to new rchttp library.

### How to review

Sanity check, run tests.

### Who can review

Anyone.

### DIFF
--- a/clients/dataset/dataset.go
+++ b/clients/dataset/dataset.go
@@ -10,9 +10,10 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/ONSdigital/dp-rchttp"
 	"github.com/ONSdigital/go-ns/clients/clientlog"
 	"github.com/ONSdigital/go-ns/common"
-	"github.com/ONSdigital/go-ns/rchttp"
+	"github.com/ONSdigital/go-ns/log"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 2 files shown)


clients/dataset/dataset.go
```diff
@@ -10,9 +10,10 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/ONSdigital/dp-rchttp"
 	"github.com/ONSdigital/go-ns/clients/clientlog"
 	"github.com/ONSdigital/go-ns/common"
-	"github.com/ONSdigital/go-ns/rchttp"
+	"github.com/ONSdigital/go-ns/log"
```
daiLlew: Is this the wrong logger? I am not sure if we can use the new logger in our go apps yet... If we can does it make sense to use PR's like this to slowly migrate across?

clients/dataset/dataset.go
```diff
@@ -88,6 +86,11 @@ func (c *Client) Get(ctx context.Context, id string) (m Model, err error) {
 
 	req = setCollectionID(ctx, req)
 
+	if len(authToken) > 0 {
```
daiLlew: I would consider moving this out into its own function which can be reused OR move the len check and the logging into `common.AddServiceTokenHeader`. That was the overhead is all in that function and you can call it where ever

clients/dataset/dataset.go
```diff
@@ -136,6 +139,11 @@ func (c *Client) GetByPath(ctx context.Context, path string) (m Model, err error
...
```
daiLlew: Like I said above you are having to repeat this multiple times in this method - feels like it makes sense to pull it out into its own thing.

clients/dataset/dataset_test.go
```diff
@@ -4,20 +4,22 @@ import (
...
 	. "github.com/smartystreets/goconvey/convey"
 	"io/ioutil"
 	"net/http"
 	"testing"
-	"github.com/ONSdigital/go-ns/common"
 )
 
 var ctx = context.Background()
 
+const authToken = "auth token"
```
daiLlew: You are going to hate me for asking this but...

We should probably assert that header is added to the request in the test....

clients/dataset/dataset.go
mikeAdamss: it's not the new-new one (i.e it's not Ians one), it's the one we used on cmd.

clients/dataset/dataset.go
mikeAdamss: simplified as per slack chat.

clients/dataset/dataset.go
mikeAdamss: simplified as per slack chat.

PR #156, opened 2019-09-04
Filter API now uses the new dp-rchttp instead of go-ns one

dandwal:
### What

Auth tokens added to the header of the request

### How to review

Check that the auth tokens are available and used correctly.

### Who can review

Anyone except me

### DIFF
--- a/clients/filter/filter.go
+++ b/clients/filter/filter.go
@@ -10,10 +10,10 @@ import (
 	"net/http"
 	"strconv"
 
+	"github.com/ONSdigital/dp-rchttp"
 	"github.com/ONSdigital/go-ns/clients/clientlog"
 	"github.com/ONSdigital/go-ns/common"
 	"github.com/ONSdigital/go-ns/log"
-	"github.com/ONSdigital/go-ns/rchttp"
 )
 
 const service = "filter-api"
@@ -50,17 +50,14 @@ var _ error = ErrInvalidFilterAPIResponse{}
 
 // Client is a filter api client which can be used to make requests to the server
 type Client struct {
-	cli common.RCHTTPClienter
+	cli rchttp.Clienter
 	url string
 }
 
 // New creates a new instance of Client with a given filter api url
-func New(filterAPIURL, serviceToken, xDownloadServiceToken string) *Client {
+func New(filterAPIURL string) *Client {
 	return &Client{
-		cli: rchttp.ClientWithDownloadServiceToken(
-			rchttp.ClientWithServiceToken(nil, serviceToken),
-			xDownloadServiceToken,
-		),
+		cli: rchttp.NewClient(),
 		url: filterAPIURL,
 	}
 }
@@ -82,7 +79,7 @@ func (c *Client) Healthcheck() (string, error) {
 }
 
 // GetOutput returns a filter output job for a given filter output id
-func (c *Client) GetOutput(ctx context.Context, filterOutputID string) (m Model, err error) {
+func (c *Client) GetOutput(ctx context.Context, filterOutputID string, authToken string, downloadServiceToken string) (m Model, err error) {
 	uri := fmt.Sprintf("%s/filter-outputs/%s", c.url, filterOutputID)
 
 	clientlog.Do(ctx, "retrieving filter output", service, uri)
@@ -91,6 +88,9 @@ func (c *Client) GetOutput(ctx context.Context, filterOutputID string) (m Model,
 	if err != nil {
 		return
 	}
+
+	common.AddServiceTokenHeader(req, authToken)
+	common.AddDownloadServiceTokenHeader(req, downloadServiceToken)
 
 	resp, err := c.cli.Do(ctx, req)
 	if err != nil {
@@ -113,7 +113,7 @@ func (c *Client) GetOutput(ctx context.Context, filterOutputID string) (m Model,
 }
 
 // GetDimension returns information on a requested dimension name for a given filterID
-func (c *Client) GetDimension(ctx context.Context, filterID, name string) (dim Dimension, err error) {
+func (c *Client) GetDimension(ctx context.Context, filterID, name string, authToken string, downloadServiceToken string) (dim Dimension, err error) {
 	uri := fmt.Sprintf("%s/filters/%s/dimensions/%s", c.url, filterID, name)
 
 	clientlog.Do(ctx, "retrieving dimension information", service, uri)
@@ -122,6 +122,9 @@ func (c *Client) GetDimension(ctx context.Context, filterID, name string) (dim D
 	if err != nil {
 		return
 	}
+
+	common.AddServiceTokenHeader(req, authToken)
+	common.AddDownloadServiceTokenHeader(req, downloadServiceToken)
 
 	resp, err := c.cli.Do(ctx, req)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


clients/filter/filter.go
```diff
@@ -92,11 +89,16 @@ func (c *Client) GetOutput(ctx context.Context, filterOutputID string) (m Model,
...
 
+	common.AddServiceTokenHeader(req, authToken)
+	common.AddDownloadServiceTokenHeader(req, downloadServiceToken)
+
 	resp, err := c.cli.Do(ctx, req)
 	if err != nil {
 		return
 	}
 
+	defer resp.Body.Close()
```
daiLlew: Another minor point but `Close` can return an error which we often don't bother to handle (again not something you've specifically done - something littered all across our apps).

I would suggest deferring a call to a close response func which logs the error:

```
func closeResponseBody(ctx context.Context, resp *http.Response) {
	if err := resp.Body.Close(); err != nil {
		log.Event(ctx, "error closing http response body", log.Error(err))
	}
}
```

```
defer closeResponseBody(ctx, resp) 
```

dandwal: Are we wanting to change the logging library for this whilst we are at it? As `clientlog` uses the go-ns logs. So by changing the logs of this file to the `log.go` logs used in the example above we can't give clientlog `log.Data` anymore. 

Alternatively, we can cast from one log form to the other and keep the old log import in the file and add a new one, but call the old one by a new keyword e.g. legacyLog instead (Not great keeping both logs, but a stepping stone in replacing the old one).

If neither then will use the old logging system for the example above.

My vote would be for us to replace the logging system but would need to agree as a team I think (as it is a bit of an increase in feature scope).

clients/filter/filter.go
```diff
@@ -92,11 +96,16 @@ func (c *Client) GetOutput(ctx context.Context, filterOutputID string) (m Model,
 		return
 	}
 
+	common.AddServiceTokenHeader(req, serviceAuthToken)
```
daiLlew: Same her: You need to set the provided context on the request otherwise it will be created with it's own which will not have any of the traceID's and other good stuff we store in it.

```
req = req.WithContext(ctx)
```

clients/filter/filter.go
```diff
@@ -123,11 +131,15 @@ func (c *Client) GetDimension(ctx context.Context, filterID, name string) (dim D
...
+
```
daiLlew: Same here

clients/filter/filter.go
```diff
@@ -159,11 +170,15 @@ func (c *Client) GetDimensions(ctx context.Context, filterID string) (dims []Dim
...
```
daiLlew: and here

clients/filter/filter.go
```diff
@@ -190,11 +204,15 @@ func (c *Client) GetDimensionOptions(ctx context.Context, filterID, name string)
...
```
daiLlew: and here

clients/filter/filter.go
```diff
@@ -246,15 +263,19 @@ func (c *Client) CreateBlueprint(ctx context.Context, datasetID, edition, versio
 		return "", err
 	}
 
+	common.AddServiceTokenHeader(req, serviceAuthToken)
+	common.AddDownloadServiceTokenHeader(req, downloadServiceToken)
+
```
daiLlew: and here

clients/filter/filter.go
```diff
@@ -291,15 +312,19 @@ func (c *Client) UpdateBlueprint(ctx context.Context, m Model, doSubmit bool) (m
 		return
 	}
 
+	common.AddServiceTokenHeader(req, serviceAuthToken)
+	common.AddDownloadServiceTokenHeader(req, downloadServiceToken)
+
```
daiLlew: and here

clients/filter/filter.go
```diff
@@ -82,7 +86,7 @@ func (c *Client) Healthcheck() (string, error) {
 }
 
 // GetOutput returns a filter output job for a given filter output id
-func (c *Client) GetOutput(ctx context.Context, filterOutputID string) (m Model, err error) {
+func (c *Client) GetOutput(ctx context.Context, filterOutputID string, serviceAuthToken string, downloadServiceToken string) (m Model, err error) {
```
daiLlew: I personally dislike the _named return parameters_. I think they are misleading and can lead to var shadowing which can be a royal pain to workout if you have a defect.
I appreciate this is not your change but I'd prefer to update this and have implicit return values.

clients/filter/filter.go
```diff
@@ -106,14 +115,13 @@ func (c *Client) GetOutput(ctx context.Context, filterOutputID string) (m Model,
...
 	}
-	defer resp.Body.Close()
 
 	err = json.Unmarshal(b, &m)
 	return
 }
 
 // GetDimension returns information on a requested dimension name for a given filterID
-func (c *Client) GetDimension(ctx context.Context, filterID, name string) (dim Dimension, err error) {
+func (c *Client) GetDimension(ctx context.Context, filterID, name string, serviceAuthToken string, downloadServiceToken string) (dim Dimension, err error) {
```
daiLlew: Same here - named return and param order

clients/filter/filter.go
daiLlew: Consistency comment: I think the param order for all client methods should be 
`ctx, serviceAuthToken, ...` followed by whatever else is needed.

clients/filter/filter.go
```diff
@@ -327,6 +352,8 @@ func (c *Client) AddDimensionValue(ctx context.Context, filterID, name, value st
 		return err
 	}
 
+	common.AddServiceTokenHeader(req, serviceAuthToken)
+
```
daiLlew: This method is missing a `defer CloseResponseBody`

clients/filter/filter.go
```diff
@@ -352,6 +379,8 @@ func (c *Client) RemoveDimensionValue(ctx context.Context, filterID, name, value
 		"value":  value,
 	})
 
+	common.AddServiceTokenHeader(req, serviceAuthToken)
+
```
daiLlew: This method is missing a `defer CloseResponseBody`

clients/filter/filter.go
```diff
@@ -377,6 +406,8 @@ func (c *Client) RemoveDimension(ctx context.Context, filterID, name string) (er
 		return
 	}
 
+	common.AddServiceTokenHeader(req, serviceAuthToken)
+
```
daiLlew: This method is missing a `defer CloseResponseBody`

clients/filter/filter.go
```diff
@@ -403,6 +434,8 @@ func (c *Client) AddDimension(ctx context.Context, id, name string) error {
 		return err
 	}
 
+	common.AddServiceTokenHeader(req, serviceAuthToken)
+
```
daiLlew: This method is missing a `defer CloseResponseBody`

clients/filter/filter.go
```diff
@@ -440,14 +476,14 @@ func (c *Client) GetJobState(ctx context.Context, filterID string) (m Model, err
 	if err != nil {
 		return
 	}
-	defer resp.Body.Close()
+	defer CloseResponseBody(ctx, resp)
```
daiLlew: The `defer` needs to be immediately after  the ` c.cli.Do` err check.

clients/filter/filter.go
```diff
@@ -471,6 +507,8 @@ func (c *Client) AddDimensionValues(ctx context.Context, filterID, name string,
 		return err
 	}
 
+	common.AddServiceTokenHeader(req, serviceAuthToken)
+
```
daiLlew: This method is missing a `defer CloseResponseBody`

clients/filter/filter.go
```diff
@@ -510,7 +551,7 @@ func (c *Client) GetPreview(ctx context.Context, filterOutputID string) (p Previ
 	if err != nil {
 		return
 	}
-	defer resp.Body.Close()
+	defer CloseResponseBody(ctx, resp)
```
daiLlew: This needs to come immediately after 
```
resp, err := c.cli.Do(ctx, req)
if err != nil {
    return
}
```

daiLlew: The token stuff is all fine but there are a few methods that are either not deferring the resp.Body.Close call at all or they are not doing immediately after checking the resp err.

daiLlew:
The token stuff is all fine but there are a few methods that are either not deferring the resp.Body.Close call at all or they are not doing immediately after checking the resp err.

PR #157, opened 2019-09-04
Update Zebedee client with new RCHTTP client

jondewijones:
### What
Updated client to new rchttp library.

### How to review
Sanity check, run tests.

### Who can review
Anyone.

### DIFF
--- a/common/identity.go
+++ b/common/identity.go
@@ -97,6 +97,13 @@ func SetFlorenceHeader(ctx context.Context, r *http.Request) {
 	}
 }
 
+// AddFlorenceHeader sets the given user access token (florence token) token on the given request
+func AddFlorenceHeader(r *http.Request, userAccessToken string) {
+	if len(userAccessToken) > 0 {
+		r.Header.Add(FlorenceHeaderKey, userAccessToken)
+	}
+}
+
 // AddAuthHeaders sets authentication headers for request
 func AddAuthHeaders(ctx context.Context, r *http.Request, serviceToken string) {
 	if IsUserPresent(ctx) {
--- a/zebedee/client/client.go
+++ b/zebedee/client/client.go
@@ -13,16 +13,16 @@ import (
 	"sync"
 	"time"
 
+	rchttp "github.com/ONSdigital/dp-rchttp"
 	"github.com/ONSdigital/go-ns/common"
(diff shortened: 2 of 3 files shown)


common/identity.go
```diff
@@ -97,6 +97,13 @@ func SetFlorenceHeader(ctx context.Context, r *http.Request) {
 	}
 }
 
+// AddFlorenceHeader sets the given user access token (florence token) token on the given request
+func AddFlorenceHeader(r *http.Request, userAccessToken string) {
```
daiLlew: I think we need to decide on a name for this and enforce consistency from here on out. For example here the method is named for `FlorenceHeader` but the parameter name is `userAccessToken`. 

I am not precious about what we call it but we need to be consistent.

zebedee/client/client.go
```diff
@@ -122,21 +122,15 @@ func (c *ZebedeeClient) GetDatasetLandingPage(ctx context.Context, path string)
...
 	}
 
-	if ctx.Value(common.FlorenceIdentityKey) != nil {
-		accessToken, ok := ctx.Value(common.FlorenceIdentityKey).(string)
-		if !ok {
-			log.ErrorCtx(ctx, errors.New("error casting access token cookie to string"), nil)
-		}
-		req.Header.Set(common.FlorenceHeaderKey, accessToken)
-	}
+	common.AddFlorenceHeader(req, userAccessToken)
```
daiLlew: See previous comment about naming.

zebedee/client/client_test.go
```diff
@@ -19,21 +21,22 @@ func TestUnitClient(t *testing.T) {
...
-	cli := NewZebedeeClient(fmt.Sprintf("http://localhost:%d", port))
-
-	ctx := context.WithValue(context.Background(), "X-Florence-Token", "test-access-token")
+	testAccessToken := "test-access-token"
+	ctx := context.Background()
 
 	Convey("test get()", t, func() {
 		Convey("test get sucessfully returns response from zebedee", func() {
-			b, err := cli.get(ctx, "/data?uri=foo")
+			b, err := cli.get(ctx, testAccessToken, "/data?uri=foo")
```
daiLlew: You are going to hate me for this _but_ I think each of these updated tests should also assert that the outbound request has the header set.

PR #159, opened 2019-09-05
created type for user and service auth tokens

daiLlew:
Proposal for auth token type. Tests to follow if accepted.

### DIFF
--- a/clients/auth_tokens.go
+++ b/clients/auth_tokens.go
@@ -1,0 +1,56 @@
+package clients
+
+// AuthToken defines common behaviour for authentication tokens
+type AuthToken interface {
+	GetValue() string
+}
+
+// UserAuthToken is a type representing user authentication tokens
+type UserAuthToken struct {
+	value string
+}
+
+// ServiceAuthToken is a type representing user authentication tokens
+type ServiceAuthToken struct {
+	value string
+}
+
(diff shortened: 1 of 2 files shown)


eldeal:
Im not convinced this is actually giving us much value for the abstraction it introduces; I don’t really think the issue is complex enough to warrant this solution at this point.

If the different token types had different validation that could be applied, so that `NewUserToken()` might return an error when it was passed a service token by mistake, then I could see a point in it. Or if we had reason to think a token might be made up from multiple inputs that can be usefully pulled together in a struct, so that `GetValue()` had to build something, then it would be more useful.

In our current situation though, they’re simple strings. We’re adding to the boilerplate an app has to call before using the client (decreasing the value of the client, imo), and it still doesn’t prevent at an application level the wrong token being passed into the wrong `NewThing()` func. I'm not sure the types are sufficiently different that type safety between them is meaningful.

I would be much more comfortable if we stuck with trying the simple solution of clearer variable names and putting more effort into the function comments before introducing this level of abstraction. If our situation changes and our tokens become more complex than simple strings, its an idea worth revisiting, but I don’t believe that’s where we are now.

_PS. if we go forward with this, or revisit it in the future, I'd ***strongly*** object to it being added to the `common` package, which I think we should be moving away from rather than adding to, especially as this is relatively well isolated with a defined subject area._

PR #160, opened 2019-09-06
Switch Hierarchy api to use the newer RCHTTP client

dandwal:
### What

1. Hierarchy-api updated to use the new RCHTTP library instead of the go-ns RHTTP library. 
2. Tests added
3. Fix issues with defer not being called to close response body at the correct time and handling errors for the closing of the response body

### How to review

1. Checkout the branch
2. Run the tests
3. Ensure that the rchttp is used correctly and that errors are handled correctly

_Note: These tests are fairly minimal as they are not technically in scope for this task - ideally they should have been there already._

### Who can review

Anyone except me

### DIFF
--- a/clients/hierarchy/hierarchy.go
+++ b/clients/hierarchy/hierarchy.go
@@ -7,9 +7,9 @@ import (
 	"io/ioutil"
 	"net/http"
 
+	"github.com/ONSdigital/dp-rchttp"
 	"github.com/ONSdigital/go-ns/clients/clientlog"
 	"github.com/ONSdigital/go-ns/log"
-	"github.com/ONSdigital/go-ns/rhttp"
 )
 
 const service = "hierarchy-api"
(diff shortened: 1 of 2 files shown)


clients/hierarchy/hierarchy.go
```diff
@@ -67,38 +75,46 @@ func (c *Client) Healthcheck() (string, error) {
...
-func (c *Client) getHierarchy(path string) (m Model, err error) {
-	resp, err := c.cli.Get(c.url + path)
+func (c *Client) getHierarchy(path string, ctx context.Context) (m Model, err error) {
+
+	req, err := http.NewRequest("GET", c.url + path, nil)
 	if err != nil {
 		return
 	}
 
+	resp, err := c.cli.Do(ctx, req)
```
daiLlew: I think you need to set the context on the request otherwise it will have its own which not contain any traceID's previously set.

```
req = req.WithContext(ctx)
```

clients/hierarchy/hierarchy.go
```diff
@@ -67,38 +75,46 @@ func (c *Client) Healthcheck() (string, error) {
...
 		"code":        code,
 	})
 
-	return c.getHierarchy(path)
+	return c.getHierarchy(path, ctx)
 }
 
-func (c *Client) getHierarchy(path string) (m Model, err error) {
-	resp, err := c.cli.Get(c.url + path)
+func (c *Client) getHierarchy(path string, ctx context.Context) (m Model, err error) {
```
daiLlew: I am personally not a fan of _named_ return parameters - I think it can be confusing and you can end up with defects because variables have been shadowed and its really hard to spot. 

I won't make you make this change to approve the request but I think it would be better if we did.

clients/hierarchy/hierarchy.go
daiLlew: Ignore this - this is taken care of inside of `clienter.Do`

PR #162, opened 2019-09-09
Feature/codelist client rchttp update

daiLlew:
Updated codelist client:

 - Replaced clienter with `dp-rchttpclient`
- Added `context` and `serviceAuthToken` parameters to all public methods.
- Added unit tests for all public methods.

### DIFF
--- a/clients/codelist/codelist.go
+++ b/clients/codelist/codelist.go
@@ -7,6 +7,7 @@ import (
 	rchttp "github.com/ONSdigital/dp-rchttp"
 	"github.com/ONSdigital/go-ns/clients/clientlog"
 	"github.com/ONSdigital/go-ns/common"
+	"github.com/ONSdigital/log.go/log"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -106,7 +107,40 @@ func (c *Client) GetValues(ctx context.Context, authToken string, id string) (Di
 }
 
 // GetIDNameMap returns dimension values in the form of an id name map
-func (c *Client) GetIDNameMap(id string) (map[string]string, error) {
+func (c *Client) GetIDNameMap(ctx context.Context, id string, authToken string) (map[string]string, error) {
+	uri := fmt.Sprintf("%s/code-lists/%s/codes", c.url, id)
+
+	resp, err := c.doWithAuthToken(ctx, "GET", uri, authToken, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			log.Event(ctx, "error closing codelistClient.GetIDNameMap response body", log.Error(err))
+		}
+	}()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	var vals DimensionValues
+	if err = json.Unmarshal(body, &vals); err != nil {
+		return nil, err
+	}
+
+	idNames := make(map[string]string)
+	for _, val := range vals.Items {
+		idNames[val.ID] = val.Label
+	}
+
+	return idNames, nil
+}
+
+// GetIDNameMap returns dimension values in the form of an id name map
+/*func (c *Client) GetIDNameMap(id string) (map[string]string, error) {
 	uri := fmt.Sprintf("%s/code-lists/%s/codes", c.url, id)
 	resp, err := c.cli.Get(context.Background(), uri)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


clients/codelist/codelist_test.go
```diff
@@ -0,0 +1,1358 @@
...
+			})
+
+			Convey("and the response body is closed", func() {
+				So(body.IsClosed, ShouldBeTrue)
+			})
+		})
+	})
+
+	Convey("given ioutil.ReadAll returns an error", t, func() {
+		expectedErr := errors.New("Peace sells, but who's buying?")
```
dandwal: I think convention in GO's error package is to not use capitals or punctuation as per https://github.com/golang/go/wiki/CodeReviewComments#error-strings
Found on lines:
1. 477
2. 597
3. 670
4. 787
5. 858
6. 975
7. 1166
3. 1166

clients/codelist/codelist.go
```diff
@@ -66,52 +67,57 @@ func (c *Client) Healthcheck() (string, error) {
 }
 
 // GetValues returns dimension values from the codelist api
-func (c *Client) GetValues(id string) (vals DimensionValues, err error) {
+func (c *Client) GetValues(ctx context.Context, serviceAuthToken string, id string) (DimensionValues, error) {
```
dandwal: Some function expect serviceAuthToken to be the last parameter whilst others like this one expect it to be after context or elsewhere in the list of parameters, might be worth us setting serviceAuthToken as the last parameter on:

1. GetValues
2. GetCodeListEditions
3. GetCodes
4. GetCodesByID
5. GetDatasetsByCode
6. doServiceRequest

Alternatively, we can have it always after context and change it in the other APIs to expected after context rather than auth tokens being the last parameters.

clients/codelist/codelist.go
```diff
@@ -124,107 +130,165 @@ func (c *Client) GetIDNameMap(id string) (map[string]string, error) {
...
 	if err != nil {
-		return
+		return code, err
 	}
+
 	return code, nil
 }
 
-func (c *Client) GetDatasetsByCode(codeListID string, edition string, codeID string) (datasets DatasetsResult, err error) {
+// GetDatasetsByCode todo
```
dandwal: 'todo'

clients/codelist/codelist.go
```diff
@@ -124,107 +130,165 @@ func (c *Client) GetIDNameMap(id string) (map[string]string, error) {
...
 	if err != nil {
-		return
+		return datasets, err
 	}
 	return datasets, nil
 }
+
+// doServiceRequest executes clienter.Do setting the service authentication token as a request header. Returns the http.Response and any error.
+// It is the callers responsibility to ensure response.Body is closed on completion.
+func (c *Client) doServiceRequest(ctx context.Context, serviceAuthToken string, method string, uri string, body io.Reader) (*http.Response, error) {
```
dandwal: I really like the use of these helper functions especially the `closeResponseBody` I think the other APIs would benefit from these too. instead of us adding these functions to each go-ns/client package it might be worth us creating some utilities as part of the dp-rchttp package which the other APIs can then use?

Having said that the doServiceRequest to me sounds like a function that just 'sends a service request' as per the name. However, technically the function is manipulating the request and not just sending it, so the function is doing more than it says it is in the name (although function annotation does state it will do this).

clients/codelist/codelist_test.go
```diff
@@ -0,0 +1,1358 @@
...
+		actual, err := codelistClient.GetValues(nil, testServiceAuthToken, "999")
+
+		So(err, ShouldBeNil)
+		So(actual, ShouldResemble, testDimensionValues)
+
+		calls := clienter.DoCalls()
+		So(calls, ShouldHaveLength, 1)
+
+		req := calls[0].Req
+		So(req.URL.Path, ShouldEqual, "/code-lists/999/codes")
```
dandwal: So this:			

```
                                calls := clienter.DoCalls()
				So(calls, ShouldHaveLength, 1)

				req := calls[0].Req
				So(req.URL.Path, ShouldEqual, uri)
				So(req.URL.Host, ShouldEqual, host)
				So(req.Method, ShouldEqual, "GET")
				So(req.Body, ShouldBeNil)
				So(req.Header.Get(common.AuthHeaderKey), ShouldEqual, common.BearerPrefix+testServiceAuthToken)
```

Is repeated 25 times in the package, wondering if we can move this into its own helper function?

clients/codelist/codelist.go
daiLlew: I think each method should have params order `ctx, serviceAuthToken......`

clients/codelist/codelist.go
daiLlew: I would counter argue that the request is not exposed to the caller - it is scoped within the function so its valid for the function do manipulate it however it needs to.

But I have renamed it to `doRequestWithServiceAuthHeader` for further transparency.

PR #168, opened 2019-10-17
refactored get auth token logic to account for cookie not found err a…

daiLlew:
- Updated `getFlorenceToken` to use new `headers.GetUserAuthToken` func.
 - Updated get cookie logic to account for cookie not found error.
 - Updated unit tests and added to coverage.
 - Updated identity handle func to return status 500 if get token returns an error.
 - Refactored identity client to allow get user/service token functions to be passed in as arguments to enable unit testing of added functionality.


identity/identity.go
```diff
@@ -23,7 +24,7 @@ func HandlerForHTTPClient(cli *clientsidentity.Client) func(http.Handler) http.H
 			ctx := req.Context()
 			log.Event(ctx, "executing identity check middleware")
 
-			florenceToken := getFlorenceTokenFromRequest(ctx, req)
+			florenceToken := getFlorenceToken(ctx, req)
 			serviceAuthToken := req.Header.Get(common.AuthHeaderKey)
```
jessjenkins: As discussed, earlier, this could also be changed from `common.` to the newer `headers.` style.

PR #169, opened 2019-10-23
Introduce permitted hack fix to add an hour to time

dandwal:
### What

Time on publishing in BST is saving as an hour earlier than it should, resulting in release date showing as day before. This PR introduces a hack fix that adds an hour on if the time is 23:00

### How to review

Checkout PR check out a legacy dataset page release date with CMD flags enabled

### Who can review

Anyone except me

### DIFF
--- a/zebedee/zebedeeMapper/mapper.go
+++ b/zebedee/zebedeeMapper/mapper.go
@@ -60,12 +60,11 @@ func MapZebedeeDatasetLandingPageToFrontendModel(dlp data.DatasetLandingPage, bc
 			sdlp.DatasetLandingPage.ReleaseDate = dlp.Description.ReleaseDate
 		}
 		sdlp.DatasetLandingPage.ReleaseDate = releaseDateInTimeFormat.Add(1 * time.Hour).Format("02 January 2006")
-
 	} else {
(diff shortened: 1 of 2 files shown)


zebedee/zebedeeMapper/mapper.go
```diff
@@ -50,8 +51,20 @@ func MapZebedeeDatasetLandingPageToFrontendModel(dlp data.DatasetLandingPage, bc
...
+		releaseDateInTimeFormat, err := time.Parse(time.RFC3339, dlp.Description.ReleaseDate)
+		if err != nil{
+			log.Error(err, nil)
+		}
+		sdlp.DatasetLandingPage.ReleaseDate = releaseDateInTimeFormat.Add(1 * time.Hour).Format("02 January 2006")
+
+	} else {
+		sdlp.DatasetLandingPage.ReleaseDate = dlp.Description.ReleaseDate
+	}
+	// END of hack fix	sdlp.DatasetLandingPage.NextRelease = dlp.Description.NextRelease
```
jondewijones: Don't think `sdlp.DatasetLandingPage.NextRelease = dlp.Description.NextRelease` be commented out?

PR #172, opened 2020-02-04
Remove deprecated libraries for data stores, queues, health checks and logging

janderson2:
### What

* The health check library has been moved and updated inline with the
latest spec.  The new version can be found in the
[dp-healthcheck repo](https://github.com/ONSdigital/dp-healthcheck).
* The data stores and queue clients have been moved to new repos:
  * [dp-s3](https://github.com/ONSdigital/dp-s3)
  * [dp-vault](https://github.com/ONSdigital/dp-vault)
  * [dp-elasticsearch](https://github.com/ONSdigital/dp-elasticsearch)
  * [dp-mongodb](https://github.com/ONSdigital/dp-mongodb)
  * [dp-kafka](https://github.com/ONSdigital/dp-kafka)
  * [dp-graph](https://github.com/ONSdigital/dp-graph) (neo4j)
* A new logging library has been written in line with the new logging spec and can be found in [log.go](https://github.com/ONSdigital/log.go).
* Also converted the repo to use go modules

### Who can review

Anyone but me.


nshumoogum:
I don't know whether you wanted to upgrade go-ns library to use go modules?

Also travis ci has failed

janderson2:
I was going to remove `go-ns/log` in a separate PR so as not to overcomplicate this one.  Since it sounds like you wouldn't mind reviewing it all together I will push those changes to this branch 😝

go.mod
```diff
@@ -0,0 +1,18 @@
+module github.com/ONSdigital/go-ns
+
+go 1.13
```
nshumoogum: I think this should be set to 1.12 for now, but generally we should be upgrading to the latest version of go (Im actually using 1.13.7)

janderson2: With the move to modules we are moving to 1.13 anyway so this library should only be used by apps built with 1.13.  It therefore seems sensible to leave this as 1.13.

PR #178, opened 2020-05-11
Updated README with deprecation notes, and deleted packages

DavidSubiros:
### What

- Removed `common`, `server` and `handlers` packages, as they have been migrated to `dp-net`
- Updated deprecation notes in README accordingly.

### How to review

- Make sure deprecation notes are clear

### Who can review

Anyone


DavidSubiros:
We should not remove packages yet.

PR #187, opened 2021-09-16
remove redundant .travis.yml as travis no longer in use

VivianAllen:
### What

removed .travis.yml as we don't use travis anywmore and SOC don't like the config being there

### How to review

check I have removed .travis.yml

### Who can review

anyone


redhug1 (approved):
LGTM

